PR #2, opened 2019-07-09
Enhance Map and Footer styleguidist examples

edellucien:


mhaertwig:
@edellucien We don't need this extra documentation. Please remove any additional documentation of components you added. Thanks.

edellucien:
Now it is clear was is required. Sorry for the unnecessary doc. Here is the fixed MR.

PR #4, opened 2019-07-09
Add lint-staged and husky for dev efficiency

edellucien:


.huskyrc.json
```diff
@@ -0,0 +1,8 @@
+{
+  "hooks": {
+    "post-checkout": "yarn install --frozen-lockfile",
```
friedjoff: Maybe use HUSKY_GIT_PARAMS to check if git is checking out a branch (see https://youtu.be/ikn_dBSski8?t=780 for details)?

edellucien: Better keep consistent the way we use those dep between our project. I include it in the package.json then :+1:

.lintstagedrc.json
```diff
@@ -1,9 +0,0 @@
-{
-  "src/**/*.js": [
-    "eslint --fix",
-    "prettier --write",
-    "git add",
-    "yarn test --bail --findRelatedTests"
-  ],
-  "src/**/*.{css,scss}": ["stylelint --fix", "git add"]
```
friedjoff: Did you remove the stylelint on purpose?

.huskyrc.json
friedjoff: Good idea!

oterral:
@edellucien  can you fix the conflict please

PR #17, opened 2019-07-25
Make zone styling configurable in CASA ZoneLayer.

mhaertwig:


src/apps/Casa/README.md
```diff
@@ -18,6 +18,13 @@ const CasaExample = () => {
   const zoneLayer = new ZoneLayer({
     // Demo token. Please replace with your own token.
     token: '5cc87b12d7c5370001c1d6551c1d597442444f8f8adc27fefe2f6b93',
+    zoneStyleFunction: (zoneProps, isSelected) => {
```
friedjoff: Maybe return the style object directly like this:
```suggestion
    zoneStyleFunction: (zoneProps, isSelected) => ({ fill: ....
```

mhaertwig: This was on purpose since I was not sure, if everybody is familiar with this Syntax.
It's the `README`.

PR #20, opened 2019-07-29
Renaming the parameter 'token' with 'apiKey'.

mhaertwig:


src/layers/TrafimageRasterLayer/TrafimageRasterLayer.js
```diff
@@ -7,38 +7,38 @@ import Layer from 'react-spatial/Layer';
...
 
-    // Add token
-    this.setToken(this.token);
+    // Add apiKey
+    this.setToken(this.apiKey);
   }
 
-  setToken(token) {
-    if (!token) {
+  setToken(apiKey) {
```
zvyn:
```suggestion
  setApiKey(apiKey) {
```

PR #40, opened 2019-08-21
Add Punctuality layer

edellucien:


public/lang/locale-de-main.json
```diff
@@ -239,5 +239,10 @@
     "Bahnhofpläne": "Bahnhofpläne",
     "Printprodukte": "Printprodukte",
     "Plakat": "Plakat",
-    "Format A4": "Format A4"
+    "Format A4": "Format A4",
+
```
oterral: Why there is spaces ?

oterral:
@edellucien can you remove spaces in locales files.

public/lang/locale-de-main.json
edellucien: IT was for categorization of translations, but we will organize better this files later.

PR #42, opened 2019-08-22
Add all the puktlichkeit layers

oterral:
Also clean package.json and local files.


package.json
```diff
@@ -1,11 +1,13 @@
 {
   "name": "trafimage-maps",
-  "proxy": "http://127.0.0.1:8000",
+  "version": "0.0.1",
```
edellucien: version '0.3.0' is already published on npm

edellucien:
```suggestion
  "version": "0.3.0",
```

PR #44, opened 2019-08-23
Fix all builds for ie and some yarn install warnings

oterral:


src/components/Menu/Menu.js
```diff
@@ -124,7 +124,7 @@ class Menu extends Component {
         </div>
 
         {loadedMenuComponents.map(Comp => (
-          <React.Suspense fallback="Loading menu...">
+          <React.Suspense fallback="Loading menu..." key={Comp.toString()}>
```
edellucien:
```suggestion
          <React.Suspense fallback="Loading menu...">
```

PR #49, opened 2019-08-26
Feature info based on promises.

mhaertwig:
Popup only for visible layers.

### DIFF
--- a/src/components/TopicLoader/TopicLoader.js
+++ b/src/components/TopicLoader/TopicLoader.js
@@ -69,7 +69,9 @@ class TopicLoader extends Component {
         .map(l => l.getFeatureInfoAtCoordinate(e.coordinate));
 
       Promise.all(infoPromises).then(featureInfos => {
-        const info = featureInfos.find(i => i.features && i.features.length);
+        const info = featureInfos
+          .reverse()
+          .find(i => i.features && i.features.length);
         dispatchSetClickedFeatureInfo(info ? { ...info } : null);
       });
     });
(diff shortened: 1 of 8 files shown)


src/components/TopicLoader/TopicLoader.js
```diff
@@ -57,7 +57,21 @@ class TopicLoader extends Component {
   }
 
   componentDidMount() {
+    const { dispatchSetClickedFeatureInfo, layerService, map } = this.props;
+
     this.updateLayers(this.topic.layers);
+
+    map.on('singleclick', e => {
```
oterral: Please remove the event on unmout with unByKey

src/components/TopicLoader/TopicLoader.js
```diff
@@ -105,13 +105,6 @@ class TopicLoader extends Component {
       if (apiKey && flatLayers[i] instanceof TrafimageRasterLayer) {
         flatLayers[i].setApiKey(apiKey);
       }
-
-      if (
-        flatLayers[i] instanceof VectorLayer ||
-        flatLayers[i] instanceof WMSLayer
-      ) {
-        flatLayers[i].onClick(this.onClick.bind(this));
```
oterral: it seems we add  the same callbacks over and over without removing the old ones.

oterral: pff forget these lines are removed not added :D

mhaertwig:
@oterral Thanks for the feedback. I addes some small commits and applied the requested changes.

PR #52, opened 2019-08-27
Add layer info button

oterral:


src/components/LayerInfosDialog/__snapshots__/LayerInfosDialog.test.js.snap
```diff
@@ -0,0 +1,5 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`LayerInfosDialog should match snapshot when Layer is defined 1`] = `null`;
```
edellucien: snapshot always empty currently. Better use a real layer for testing?

oterral: that's the point. I want it returns null if there is no layer.

oterral: Tests fixed

edellucien: Thanks :)

src/components/Dialog/Dialog.js
```diff
@@ -0,0 +1,48 @@
...
+  footer: PropTypes.element,
+};
+
+const defaultProps = {
+  title: null,
+  body: null,
+  footer: null,
+};
+
+const hideDialog = dispatch => () => {
```
mhaertwig: @oterral do we really need this function to call a function with a function? :wink:

oterral: ahah not anymore indeed. Thx for pointing this out

PR #53, opened 2019-08-27
Lucien/deselect train tracker

edellucien:


src/menus/TrackerMenu/TrackerMenu.js
```diff
@@ -52,9 +52,11 @@ class TrackerMenu extends Component {
         );
 
         layer.onClick(traj => {
-          dispatchSetMenuOpen(false);
+          if (traj) {
+            dispatchSetMenuOpen(false);
+          }
           this.setState({
-            open: layer.getName(),
+            open: layer.getName() || false,
```
oterral: layer.getName() can't be null I think

PR #56, opened 2019-08-28
Lucien/more fixes

edellucien:


src/components/Permalink/Permalink.js
```diff
@@ -36,6 +36,10 @@ const LINE_FILTER = 'PublishedLineName';
 const ROUTE_FILTER = 'TripNumber';
 const OPERATOR_FILTER = 'Operator';
 
+const getUrlParam = param => {
+  return param ? decodeURI(param).replace(/\s+/g, '') : undefined;
```
oterral:
```suggestion
  return decodeURI(param).replace(/\s+/g, '');
```

src/components/Permalink/Permalink.js
```diff
@@ -36,6 +36,10 @@ const LINE_FILTER = 'PublishedLineName';
 const ROUTE_FILTER = 'TripNumber';
 const OPERATOR_FILTER = 'Operator';
 
+const getUrlParam = param => {
```
oterral:
```suggestion
const getUrlParam = (param = '') => {
```

src/components/Permalink/Permalink.js
```diff
@@ -63,9 +67,9 @@ class Permalink extends PureComponent {
       dispatchSetZoom(z);
     }
 
-    const lineFilter = parameters[LINE_FILTER];
-    const routeFilter = parameters[ROUTE_FILTER];
-    const operatorFilter = parameters[OPERATOR_FILTER];
+    const lineFilter = getUrlParam(parameters[LINE_FILTER]);
+    const routeFilter = getUrlParam(parameters[ROUTE_FILTER]);
+    const operatorFilter = getUrlParam(parameters[OPERATOR_FILTER]);
```
oterral: You don' manage the case insentivity

oterral:
```suggestion
    const operatorFilter = getUrlParam(parameters[OPERATOR_FILTER.toLowerCase()]);
```

src/components/Permalink/Permalink.js
oterral: In fatc there is a decode option for qs.parse .https://www.npmjs.com/package/query-string#decode

src/components/Permalink/Permalink.js
edellucien: if the param is not undefined, it will be set as empty in the url. That is not what we need.

src/components/Permalink/Permalink.js
edellucien: works as expected with the current version: 
'publishedlinename=s1, s2, S 15, S  9'

edellucien: But needs this PR to be merged first: https://github.com/geops/react-transit/pull/34

oterral:
PR in react-transit merged can I merge this one too or I wait ?

edellucien:
It is resolved yes

PR #57, opened 2019-08-28
Timeout after menu close, for calculating Tracker Menu max height

edellucien:


oterral:
I'm pretty sure if there is 2 Menu item it will not work . But good for know.

PR #58, opened 2019-08-28
Eva/container

gevos:


zvyn:
LGTM

src/config/containerTopics.js
```diff
@@ -0,0 +1,10 @@
+import { netzkarteLayer } from './layers';
```
mhaertwig: Please see ticket notes.

PR #64, opened 2019-08-30
Passagierfrequenzen

gevos:


src/components/Popup/Popup.js
```diff
@@ -45,7 +45,7 @@ const Popup = ({
 
   // Styleguidist try to load every file in the folder if we don't put index.js
   const PopupComponent = React.lazy(() =>
-    import(`../../popups/${componentName}/index.js`),
+    import(`../Popups/${componentName}/index.js`),
```
mhaertwig: @edellucien why did you move this directory? I would like to have `menus` and `popups` next to `components`.

src/layers/PassagierfrequenzenLayer/PassagierfrequenzenLayer.js
```diff
@@ -0,0 +1,190 @@
...
+
+class PassagierfrequenzenLayer extends VectorLayer {
+  constructor(options = {}) {
+    const name = 'ch.sbb.bahnhoffrequenzen';
+    const key = 'ch.sbb.bahnhoffrequenzen';
+
+    const olLayer = new OLVectorLayer({
+      source: new OLVectorSource({
+        format: new OLGeoJSON(),
+        ...(options.useBboxStrategy ? { strategy: OLBboxStrategy } : {}),
```
mhaertwig: @edellucien as discussed yesterday, please request all features at once.

src/components/Popup/Popup.js
edellucien: This has been decided to be moved in a intern ticket.

edellucien: Fixed

src/config/layers.js
```diff
@@ -172,6 +173,11 @@ export const swisstopoLandeskarteGrau = new Layer({
   }),
 });
 
+export const passagierfrequenzen = new PassagierfrequenzenLayer({
+  visible: false,
+  useBboxStrategy: true,
```
mhaertwig: @edellucien is this needed?

src/layers/PassagierfrequenzenLayer/PassagierfrequenzenLayer.js
```diff
@@ -0,0 +1,184 @@
...
+      version: '1.0.0',
+      request: 'GetFeature',
+      typeName: 'trafimage:passagierfrequenzen',
+    };
+    this.olLayer.setStyle((f, r) => this.styleFunction(f, r));
+    this.olLayer.getSource().setLoader(this.loader.bind(this));
+
+    this.onClick(f => {
+      const [clickedFeature] = f;
+      this.clickedFeatureName = clickedFeature && clickedFeature.get('name');
```
mhaertwig: Please use the feature itself instead of its name.

src/layers/PassagierfrequenzenLayer/PassagierfrequenzenLayer.js
```diff
@@ -0,0 +1,184 @@
...
+    this.onPointerMoveRef = this.map.on('pointermove', e =>
+      this.onPointerMove(e),
+    );
+  }
+
+  styleFunction(feature, resolution) {
+    const vis = feature.get('visibility');
+    const res = layerHelper.getDataResolution(resolution);
+    const name = feature.get('name');
+    const selected = this.clickedFeatureName === name;
```
mhaertwig: see above

src/layers/PassagierfrequenzenLayer/PassagierfrequenzenLayer.js
```diff
@@ -0,0 +1,184 @@
...
+        this.olLayer.getSource().clear();
+        this.olLayer.getSource().addFeatures(features);
+      })
+      .catch(() => {
+        this.olLayer.getSource().removeLoadedExtent(extent);
+      });
+  }
+
+  onPointerMove(evt) {
+    if (evt.dragging) {
```
mhaertwig: we need this behaviour for all features. please integrate it in the map component and remove it from all layers.

src/layers/PassagierfrequenzenLayer/PassagierfrequenzenLayer.js
edellucien: There is a feature per resolution, the comparison feature against feature will fail if I zoom in. Therefore the highlighting only works for the same resolution, and disappear when zooming. Using the name is the solution to ensure the highlighting works when zooming

src/components/Map/Map.js
```diff
@@ -87,6 +97,21 @@ class Map extends PureComponent {
...
+  onPointerMove(evt) {
+    const { map } = this.props;
+
+    if (evt.dragging) {
+      return;
+    }
+
+    const mapFeatures = map.getFeaturesAtPixel(evt.pixel);
+
+    const newHoveredStation =
```
mhaertwig: the variable name is misleading. please use something like `hoverFeature`;

PR #66, opened 2019-09-02
Link topics

mhaertwig:
Topics that link to the old wkp.
To be removed after the migration is finished.

### DIFF
--- a/src/AppRouter.js
+++ b/src/AppRouter.js
@@ -20,7 +20,7 @@ const AppRouter = () => (
           return (
             <TrafimageMaps
               history={history}
-              activeTopicKeyopic={match.params.topic}
+              activeTopicKey={match.params.topic}
               topics={topics}
               apiKey="5cc87b12d7c5370001c1d6551c1d597442444f8f8adc27fefe2f6b93"
               elements={{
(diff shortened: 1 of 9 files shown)


src/AppRouter.js
```diff
@@ -20,7 +20,7 @@ const AppRouter = () => (
           return (
             <TrafimageMaps
               history={history}
-              activeTopicKeyopic={match.params.topic}
+              activeTopicKey={match.params.topic}
```
oterral: Nice found

src/components/TopicLoader/TopicLoader.js
```diff
@@ -53,13 +57,16 @@ class TopicLoader extends Component {
       ? TOPIC_CONF.find(t => t.key === activeTopicKey)
       : topics[0];
 
+    if (this.topic.linkUrl) {
+      TopicLoader.openLinkTopic(this.topic);
```
oterral: What are these links ? links to other websites ?

src/config/topics.js
```diff
@@ -37,4 +37,41 @@ export const casa = {
   projection: 'EPSG:3857',
 };
 
-export default [netzkarte];
+export const bauprojekte = {
+  name: 'ch.sbb.construction',
+  key: 'ch.sbb.construction',
+  linkUrl: 'https://maps.trafimage.ch/#/ch.sbb.construction',
```
oterral: Permalink should take care of changing the url when the topic change. This link  doesn't seem a robust solution.

PR #73, opened 2019-09-09
Fix footer css

oterral:
Fix select css broken by base layer select box


oterral:
I merge myself because it's ugly in prod

PR #77, opened 2019-09-10
Add footer links

oterral:


src/components/Dialog/Dialog.scss
```diff
@@ -30,6 +32,20 @@
   left: 390px;
 }
 
+.tm-modal {
```
mhaertwig: please use more explicit styler definitions instead of `!important`

oterral: Fixed witha  container

oterral:
@mhaertwig  Ready to re review

PR #79, opened 2019-09-11
Updates for CASA

mhaertwig:


oterral:
@mhaertwig casa example does not work in preview

oterral:
it worked

PR #88, opened 2019-09-13
Routelayer is clickable

mhaertwig:


src/layers/RouteLayer/RouteLayer.js
```diff
@@ -78,7 +78,7 @@ class RouteLayer extends VectorLayer {
    * @param {String} mot ask for specific Route
    * @returns {array<ol.feature>}
```
edellucien: would need to add the new param in the method documentation

PR #134, opened 2019-09-30
Fix Menu style

edellucien:


src/components/Menu/TopicMenu.js
```diff
@@ -198,29 +218,7 @@ class TopicMenu extends PureComponent {
             topic.description &&
             this.renderInfoButton(topic)}
         </div>
-        <div className="wkp-topic-content">
```
oterral: why not putting the Collapsible here instead ? just after `wkp-topic-content` , otherwise  the base layer selector will not be displayed if there is not layerTree

edellucien: Good idea! Thanks

PR #138, opened 2019-10-01
Lucien/handicap highlight

edellucien:


edellucien:
Activating your CORS header plugin to see the features is mandatory

PR #145, opened 2019-10-08
Handicap updates

mhaertwig:


friedjoff:
Looks good! The only thing I noticed was other components which still get map and layerService as props. Maybe better to handle this in antoher PR.

src/AppRouter.js
```diff
@@ -17,37 +20,34 @@ const AppRouter = () => (
       path="/:topic"
       render={({ history, match }) => {
         if (topicKeys.includes(match.params.topic)) {
+          const activeTopic = topics.find(t => t.key === match.params.topic);
+          const { elements } = activeTopic;
+          let menuChildren = null;
+
+          switch (match.params.topic) {
+            case 'ch.sbb.handicap': {
```
friedjoff: Would be nice to make AppRouter independent from configuration details like menus specific to certain topics. But currently I don't have a good idea how to do this.

src/AppRouter.js
```diff
@@ -17,37 +20,34 @@ const AppRouter = () => (
       path="/:topic"
       render={({ history, match }) => {
         if (topicKeys.includes(match.params.topic)) {
+          const activeTopic = topics.find(t => t.key === match.params.topic);
+          const { elements } = activeTopic;
```
friedjoff: Can't we extract `elements` from `topics` inside TrafimageMaps?

src/components/FeatureMenu/FeatureMenu.js
```diff
@@ -1,215 +1,54 @@
...
 FeatureMenu.propTypes = propTypes;
 FeatureMenu.defaultProps = defaultProps;
 
 export default compose(
   withTranslation(),
-  connect(
-    mapStateToProps,
-    mapDispatchToProps,
-  ),
+  connect(mapStateToProps),
```
oterral: No need of `connect` and `mapStateToProps` and  if you use hooks:
```
const clickedFeatureInfo = useSelector(state=> state.app.clickedFeatureInfo);
```

src/components/TrafimageMaps/TrafimageMaps.js
```diff
@@ -144,14 +125,19 @@ const defaultProps = {
...
   apiKey: null,
   history: null,
   initialState: {},
 };
 
+/**
+ * Application context.
+ * Only use the global app context for the map and the layerService.
+ */
+export const AppContext = React.createContext();
```
oterral: It seems a bit weird to me to use redux AND context api in the same time, putting map and layerService in redux state will have the same effect..

Even if it's one line of code I will put it in another file..

src/popups/HandicapPopup/HandicapPopupElement.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { withTranslation } from 'react-i18next';
+import { compose } from 'lodash/fp';
+
+const propTypes = {
+  t: PropTypes.func.isRequired,
+  properties: PropTypes.shape.isRequired,
+  k: PropTypes.string.isRequired,
```
oterral: What is the meaning of `k` and `t` ?

src/popups/HandicapPopup/HandicapPopupElement.js
```diff
@@ -0,0 +1,68 @@
...
+    }
+    default:
+      return null;
+  }
+}
+
+PopupElement.propTypes = propTypes;
+PopupElement.defaultProps = defaultProps;
+
+export default compose(withTranslation())(PopupElement);
```
oterral: simpler to use `useTranslation` hook, to get the t function.

src/popups/HandicapPopup/HandicapPopup.scss
```diff
@@ -0,0 +1,7 @@
+.handicap-popup {
```
oterral: can we use the `wkp-` prefix?

src/components/TrafimageMaps/TrafimageMaps.js
friedjoff: In my view it makes sense to use the context API to share an instance object with multiple components. Redux seems better suited to mutate and listen to global app state changes.

src/components/FeatureInformation/FeatureInformation.js
```diff
@@ -0,0 +1,76 @@
...
+import { IoIosArrowRoundBack, IoIosArrowRoundForward } from 'react-icons/io';
+import './FeatureInformation.scss';
+
+const propTypes = {
+  clickedFeatureInfo: PropTypes.array.isRequired,
+  popupComponents: PropTypes.objectOf(PropTypes.string).isRequired,
+  t: PropTypes.func.isRequired,
+};
+
+const FeatureInformation = ({ clickedFeatureInfo, popupComponents, t }) => {
```
oterral: you can use the `useTranslation` hook to get the `t` function

src/AppRouter.js
oterral: I agree. Couldn't we define the menuChildren and popups conf in the topic config directly?

src/AppRouter.js
```diff
@@ -17,37 +20,34 @@ const AppRouter = () => (
...
-                },
-                {
-                  component: HandicapMenu,
-                  standalone: true,
-                  topic: 'ch.sbb.handicap',
-                },
-              ]}
-            />
+            >
+              <Menu subMenus=<ShareMenu />>{menuChildren}</Menu>
```
oterral: This definitinon is a bit confusing, couldn't we have something like :
```
<Menu>
    <TopicsMenu>
        <ShareMenu/>
    </TopicsMenu>
    {menuChildren}
</Menu>
```
Or if `ShareMenu` is always there we can get rid of `subMenus` property, and put `ShareMenu` directly in `Menu`

src/AppRouter.js
mhaertwig: I also agree but I don't have a solution.
Can we define react components in the config file?
However, this problem still persists. I would leave it like this until we find a better solution. Do you agree?

src/components/TrafimageMaps/TrafimageMaps.js
mhaertwig: I asked the same question to friedjoff yesterday and agreed with his answer.

src/AppRouter.js
friedjoff: 👍

src/popups/HandicapPopup/HandicapPopupElement.js
mhaertwig: I got inspired by someone else... ;-)
I agree with you. The handicap popup is more a draft and it will definitely not stay like this. However, `k` should be renamed to `key`.

src/AppRouter.js
mhaertwig: If this can be done easily, please fix this for me, @edellucien.
Otherwise we leave it like this for now. OK?

mhaertwig:
> EDIT: And it seems the popups in the demos doesn't work as expected
@edellucien can you please have a look at this? sorry for letting you fix my bugs... :disappointed:

src/AppRouter.js
mhaertwig: Not exactly sure what you mean.

friedjoff: To me it looks like this can be moved into the TrafimageMaps component, but maybe I'm missing something.

src/AppRouter.js
oterral: ok

src/AppRouter.js
oterral: yes and that's break menu display.

oterral:
According to our discussions this morning:

-  map and layerService are created at initialisation time and stored in redux  see [here|https://github.com/geops/trafimage-maps/pull/145/files#diff-74f72a0e52a59493c31c664d9f654a0eR24-R33]
- popups and menus has been fixed.
- I have created  a TopicsMenu components to be able to define the Menu content like this:
```<Menu>
    <TopicsMenu>
        <ShareMenu/>
    </TopicsMenu>
    {menuChildren}
</Menu>```

**Comment by oterral** [comment]
@lucien  kannst du schauen bitte

File: src/AppRouter.js (hunk 2 above)
Comment (oterral) [reply src/AppRouter.js]: Menus display fixed

File: src/components/TrafimageMaps/TrafimageMaps.js [hunk 10]
```diff
@@ -144,124 +127,113 @@ const defaultProps = {
...
+  const defaultElements = {
+    header: <Header />,
+    popup: <Popup popupComponents={popupComponents} />,
+    footer: <Footer />,
+    permalink: <Permalink history={history} initialState={initialState} />,
+    mapControls: <MapControls />,
+    menu: (
+      <Menu>
+        <TopicsMenu>
+          <ShareMenu />
```
mhaertwig: :+1: That's great @oterral. Thanks!

PR #151, opened 2019-10-11
WIP: Lucien/barrierfree

edellucien:


src/AppRouter.js
```diff
@@ -3,41 +3,31 @@ import { BrowserRouter as Router, Route, Redirect } from 'react-router-dom';
...
 const AppRouter = () => (
   <Router>
     <Route exact path="/" render={() => <Redirect to={`${topicKeys[0]}`} />} />
     <Route
       exact
       path="/:topic"
-      render={({ history, match }) => {
-        if (topicKeys.includes(match.params.topic)) {
+      component={({ history, match }) => {
+        const topicKey = match.params.topic;
```
mhaertwig: using `match.params.topic` directly is more explicit (also see line 19).

src/Globals.scss
```diff
@@ -5,6 +5,10 @@ $footer-font-size: 12px;
 
 $transition-delay: 300;
 
+// BarrierFree
+$bf-large-outline: 4px solid black !important;
```
mhaertwig: Please use a blue outline.

mhaertwig:
I don't understand why we need to focus so many elements manually (footer, layer toggler, header). Please let's discuss this. The goal should be to only implement the focussing of features.

edellucien:
Close in profit from: https://github.com/geops/trafimage-maps/pull/170

PR #155, opened 2019-10-14
Search for handicap topic

friedjoff:


src/searches/StationFinder/StationFinder.js
```diff
@@ -0,0 +1,38 @@
...
+class StationFinder extends Search {
+  constructor() {
+    super();
+    this.key = '5cc87b12d7c5370001c1d6551c1d597442444f8f8adc27fefe2f6b93';
+  }
+
+  search(value) {
+    return fetch(`https://api.geops.io/stops/v1/?&q=${value}&key=${this.key}`)
+      .then(data => data.json())
+      .then(featureCollection => featureCollection.features);
```
oterral: Maybe transform directly the featureCollection to openlayers features.

friedjoff: This function runs often, so my idea was to keep this as performant as possible.

src/searches/StationFinder/StationFinder.js
```diff
@@ -0,0 +1,38 @@
...
+
+  render(item) {
+    return <div>{item.properties.name}</div>;
+  }
+
+  select(item) {
+    this.dispatch(setZoom(12));
+    this.dispatch(
+      setCenter(
+        transformCoords(item.geometry.coordinates, 'EPSG:4326', 'EPSG:3857'),
```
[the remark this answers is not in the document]
friedjoff: Good idea, I'll change this. I'm currently also working on a generic highlighting functionality.

src/components/Map/Map.js
```diff
@@ -67,6 +70,13 @@ class Map extends PureComponent {
     this.onPointerMoveRef = map.on('pointermove', e => this.onPointerMove(e));
   }
 
+  componentDidUpdate(prevProps) {
+    const { highlight } = this.props;
+    if (prevProps.highlight !== highlight) {
+      console.log('TODO: highlight!!!', highlight);
```
oterral: Pronlem is how do we know which highlight to apply. Highlight from a layer, from the search , from the app...

src/components/Search/Search.js
```diff
@@ -0,0 +1,105 @@
...
+        const newSuggestions = [...oldSuggestions];
+        if (index === -1 && items) {
+          newSuggestions.push({ section, items });
+        } else if (items) {
+          newSuggestions[index] = { section, items };
+        }
+        return newSuggestions;
+      });
+    return new SearchService(activeTopic, clear, upsert, dispatch);
+  }, [activeTopic, setSuggestions, dispatch]);
```
oterral: `setSuggestion` and `dispatch` changes on every render, aren't they?  Can you verify that the `useMemo` does it's work properly.

src/components/Map/Map.js
friedjoff: `highlight` would contain an OpenLayers feature which will be rendered in a VectorLayer with a generic highlight style. If our highlighting requirements are more complex we should propably implement this directly in the Search component and SearchService.

src/components/Search/Search.js
friedjoff: I'll check...

src/components/Search/Search.js
```diff
@@ -0,0 +1,105 @@
+import React, { useMemo, useState } from 'react';
+import Autosuggest from 'react-autosuggest';
```
oterral: Have you seen our Autocomplete component in `react-spatial` ?  I'm not against using a better autocomplete component but we  fix a lot of behavior stuff (css, keyboard navigation) and bugs that I don't want to fix them again :)

src/components/Search/Search.js
```diff
@@ -0,0 +1,105 @@
...
+    return new SearchService(activeTopic, clear, upsert, dispatch);
+  }, [activeTopic, setSuggestions, dispatch]);
+
+  return (
+    Object.keys(activeTopic.searches).length > 0 && (
+      <div className="wkp-search">
+        <SearchToggle>
+          <Autosuggest
+            multiSection
+            suggestions={suggestions}
```
oterral: How does work the multiSection ? each suggestion has a section property?

friedjoff: Suggestions are grouped by section (see https://codepen.io/moroshko/pen/qbRNjV for an example). To be more flexible I've also added the section property to the suggestion.

src/components/Search/Search.js
```diff
@@ -0,0 +1,105 @@
...
+        }
+        return newSuggestions;
+      });
+    return new SearchService(activeTopic, clear, upsert, dispatch);
+  }, [activeTopic, setSuggestions, dispatch]);
+
+  return (
+    Object.keys(activeTopic.searches).length > 0 && (
+      <div className="wkp-search">
+        <SearchToggle>
```
oterral: Do we need the searchToggle ?

src/components/Search/Search.js
friedjoff: Mario and me discussed this and we would like to remove non-spatial components from react-spatial. In my opinion we should move them into a generic UI component library or use third-party component libraries. This was my attempt on the autocomplete library. But we should discuss this topic together and decide sooner than later how to proceed.

src/components/Search/Search.js
friedjoff: I tried to replicate the behavior from the old WKP. On small mobile screens the search input field is in conflict with the SBB logo, so I guess we need some kind of toggle. But I'm open to alternatives.

src/components/Search/Search.js
oterral: I agree we should remove these components from react-spatial.

src/components/Search/Search.js
oterral: Yes exactly the kind of behavior we fixed with autocomplete. It should work with just css :
https://plans2.dev.trafimage.geops.ch/bern?c=600120.07%2C199658.77&lang=de&layer=bern_aussenplan&x=600107.38&y=199654.45&z=19

src/components/Search/Search.js
oterral: Ah yes I didn't have the last version , grouping is very nice.

src/components/Search/Search.js
friedjoff: Good point, I'll try that.

src/components/Search/Search.js
friedjoff: Seems to work, at least a new SearchService is only initilized if activeTopic changes.

src/components/TopicLoader/TopicLoader.js
```diff
@@ -83,13 +85,23 @@ class TopicLoader extends Component {
...
     if (activeTopic !== prevProps.activeTopic) {
       if (activeTopic.linkUrl) {
         TopicLoader.openLinkTopic(activeTopic);
       }
       this.updateLayers(activeTopic.layers);
+
+      searchService.setSearches(activeTopic.searches);
+      searchService.setSearchProps({
+        activeTopic,
+        dispatchSetClickedFeatureInfo,
```
oterral: since searchs meant to be generic acroos applications, I would not put dispatch function in properties. may be an 

`onItemSelect: (featureInfo) => {
  dispatchSetClickedFeatureInfo(featureInfo)
}`  
will be more appropriate.

src/components/Search/SearchService.js
```diff
@@ -0,0 +1,96 @@
...
+  setMap(map) {
+    this.map = map;
+    this.highlightLayer.setMap(map);
+  }
+
+  setSearches(searches) {
+    this.searches = searches;
+  }
+
+  setSearchProps(props) {
```
oterral: `setSearchesProps` maybe

src/searches/StationFinder/StationFinder.js
oterral: yes you're right

src/components/TopicLoader/TopicLoader.js
friedjoff: `dispatchSetClickedFeatureInfo` will only be used by `HandicapStationFinder`, so a generic `onItemSelect` callback for all searches will not work in this case. Also `setSearchProps` will  only be called by custom apps and those custom props only be consumed by custom SearchClasses specific to the app (`HandicapStationFinder` should not be moved to `react-transit`). It looks a little awkward, but it's the best way I found for this problem.

src/components/Search/SearchService.js
friedjoff: Good point, done.

src/components/TopicLoader/TopicLoader.js
oterral: Ok I see , i can live with that

PR #156, opened 2019-10-14
Abfahrtszeiten: Init popup (TransportPopup)

gevos:


src/popups/TransportPopup/TransportPopup.scss
```diff
@@ -0,0 +1,137 @@
...
+    -webkit-transform: rotate(0deg);
+  }
+
+  to {
+    -webkit-transform: rotate(360deg);
+  }
+}
+
+@keyframes spin {
+  from {
```
oterral: I think you don't need to define prefix . `autoprefixer` should do it automatically.

edellucien:
Requires react-spatial changes: https://github.com/geops/react-spatial/pull/314

PR #157, opened 2019-10-15
Fix yarn.lock file for doc building

edellucien:


edellucien:
Downgrade some deps, the preview is working.
Checked out from broken master.

PR #178, opened 2019-10-21
Some fixes

oterral:


src/components/Search/Search.url.svg
```diff
@@ -0,0 +1,7 @@
+<svg version="1.1" id="Search"
```
friedjoff: Looks like a duplicate of `./Search.svg` file. Can we reuse the existing file?

src/components/TopicLoader/TopicLoader.js
```diff
@@ -97,7 +97,7 @@ class TopicLoader extends Component {
       }
       this.updateLayers(activeTopic.layers);
 
-      searchService.setSearches(activeTopic.searches);
+      searchService.setSearches(activeTopic.searches || []);
```
friedjoff:
```suggestion
      searchService.setSearches(activeTopic.searches || {});
```
Not sure if this makes a difference, but searchService expects an object for searches.

src/components/Search/Search.url.svg
oterral: I guess we could it's just a renaming

oterral: Done

friedjoff:
Thanks!

PR #188, opened 2019-10-23
Mario/handicap updates

mhaertwig:


oterral:
sorry there is a conflict :), merge it when it's fixed

PR #199, opened 2019-10-28
Use style from tileserver

oterral:


oterral:
I megre directyl because it fix docuemntation on prod

PR #203, opened 2019-10-29
fix mobile search

friedjoff:
I switched the search SVG icon back to a React component, because it didn't show up in all browsers (could be fixed with CSS) and lost hover styling. To fix our documentation I also updated the styleguidist webpack config to be more similiar to what create-react-app is doing. @oterral can you please check if this breaks anything else?


friedjoff:
Discussed with @oterral and we decided to restore the url-loader for .url.svg icons to make sure react-spatial is supported.

PR #206, opened 2019-10-30
New handicap popup design

edellucien:


src/components/Popup/Popup.scss
```diff
@@ -6,6 +6,12 @@
     -o-animation: fadein 0.5s; /* Opera < 12.1 */
     animation: fadein 0.5s;
   }
+
+  .tm-popup-close-bt {
+    height: 40px !important;
+    width: 40px !important;
```
oterral: hmm could we avoid '!important' there ?

edellucien: If we define it above, but I thought it clearer to define it in Popus.scss, and add important to avoid to be rewritten.

mhaertwig: @edellucien more explicit rules overwrite less explicit ones.
`!important` is almost never needed.

src/popups/HandicapPopup/HandicapPopup.js
```diff
@@ -46,44 +46,74 @@ function HandicapPopup({ feature }) {
...
-      <PopupElement
-        label="Zusätzliche Informationen"
-        properties={properties}
-        propertyName={`zusaetzliche_informationen_${language}`}
-      />
+      <table>
+        {elementsList.map(field => (
+          <tr>
+            <td className="wkp-handicap-popup-field-title">{field.label}</td>
+            <td>{field.element}</td>
```
mhaertwig: what if there's no content available?

src/popups/HandicapPopup/HandicapPopup.js
```diff
@@ -46,44 +46,74 @@ function HandicapPopup({ feature }) {
...
-        label="Dienstleistungen Dritter"
-        properties={properties}
-        propertyName={`beschreibung_zur_dritte_dienstleistung_${language}`}
-      />
-      <PopupElement
-        label="Zusätzliche Informationen"
-        properties={properties}
-        propertyName={`zusaetzliche_informationen_${language}`}
-      />
+      <table>
```
mhaertwig: nah, please no table layout...
the first column should have a fixed width. otherwise it becomes too big.
see screen design.

src/components/Popup/Popup.scss
```diff
@@ -1,14 +1,21 @@
 .tm-popup {
   div:first-child {
```
mhaertwig: please let's not style div's without class names or ids.

src/popups/HandicapPopup/HandicapPopup.scss
```diff
@@ -1,10 +1,37 @@
 .wkp-handicap-popup {
   .wkp-handicap-popup-title {
+    background-color: #f5f5f5;
     font-weight: bold;
-    margin: 10px 0 5px 0;
+    position: absolute;
```
mhaertwig: why absolute positioning?

src/popups/HandicapPopup/HandicapPopup.scss
```diff
@@ -1,10 +1,37 @@
...
+  }
+
+  .wkp-handicap-popup-field-title {
+    font-weight: bold;
+    vertical-align: top;
+    padding-right: 20px;
+  }
+}
+
+.tm-w-m,
```
mhaertwig: please add the `.tm-app` (?) class.

src/popups/HandicapPopup/HandicapPopupElement.js
```diff
@@ -1,43 +1,29 @@
...
-        <div>
-          <u>{t(propLabel)}</u>
-        </div>
-        <div>
-          {values.map(v => (
-            <div>{v}</div>
-          ))}
-        </div>
-      </>
+      <div>
```
mhaertwig: is this `<div>` needed?

src/components/Popup/Popup.scss
oterral: yes, the popup css is not ideal in react-spatial that's why there is this div:first-child.

src/popups/HandicapPopup/HandicapPopup.scss
edellucien: To position this header on the top of the popup.

src/popups/HandicapPopup/HandicapPopup.js
edellucien: I updated it to only show properties that have value

src/popups/HandicapPopup/HandicapPopup.js
edellucien: I changed to display flex and fixed width of columns

src/popups/HandicapPopup/HandicapPopup.scss
oterral: Also related to the css of react-spatial-popup. Will be fixed when we clean react-spatial.

mhaertwig:
I think it's time to clean up the css of the react-spatial popup. @oterral will implement the bus popup soon and before overwriting the bad css in both projects, we should instead take a few minutes to clean it up.

edellucien:
Requirements changed.
FIx in: https://github.com/geops/trafimage-maps/pull/242

PR #223, opened 2019-11-05
Showcases: Init topic

gevos:


gevos:
As long as we don't show the base layer toggler at the same time, the integration of the base layers in the menu is no problem.

mhaertwig:
@gevos the new layers should not be integrated in the example. Moreover, the doc is broken.
Maybe it can be resolved by merging master. Otherwise it looks good for me.

gevos:
> the new layers should not be integrated in the example.

Done

> Moreover, the doc is broken.

Fixed in https://github.com/geops/trafimage-maps/pull/225

PR #240, opened 2019-11-07
Fix env variables in documentation

friedjoff:


src/components/TrafimageMaps/TrafimageMaps.js
```diff
@@ -203,10 +203,10 @@ const defaultProps = {
...
   menus: null,
   subMenus: null,
-  cartaroUrl: null,
-  geoServerUrl: null,
-  vectorTilesKey: null,
-  vectorTilesUrl: null,
+  cartaroUrl: process.env.REACT_APP_CARTARO_URL,
+  geoServerUrl: process.env.REACT_APP_GEOSERVER_URL,
+  vectorTilesKey: process.env.REACT_APP_VECTOR_TILES_KEY,
+  vectorTilesUrl: process.env.REACT_APP_VECTOR_TILES_URL,
```
oterral: Can't we put prod urls by default? Don't forget  the app and the web-compoent will be separate. So these variables will not exist.

oterral: or may be they will also exits in the web component build.

src/AppRouter.js
```diff
@@ -48,10 +48,6 @@ const AppRouter = () => (
             elements={elements}
             initialState={{ ...match.params }}
             popupComponents={POPUP_CONF}
-            cartaroUrl={process.env.REACT_APP_CARTARO_URL}
-            geoServerUrl={process.env.REACT_APP_GEOSERVER_URL}
-            vectorTilesKey={process.env.REACT_APP_VECTOR_TILES_KEY}
-            vectorTilesUrl={process.env.REACT_APP_VECTOR_TILES_URL}
```
oterral: We will still need this Ithink when the AppRouter will exits only in gitlab

friedjoff: Yeah, we can add them again, if needed. Just wanted to move the default value into TrafimageMaps.

src/components/TrafimageMaps/TrafimageMaps.js
friedjoff: THE web component would need to support the .env file. In that case it's the same as putting prod urls by default in here.

oterral: yes

PR #245, opened 2019-11-08
Fix click on next and previous buttons

oterral:


edellucien:
Thanks for fixing :+1:

PR #246, opened 2019-11-08
Mario/webcomponent config

mhaertwig:


mhaertwig:
@oterral some cleanup

oterral:
nice thx

PR #249, opened 2019-11-08
Gevos/construction

gevos:


src/components/FeatureInformation/FeatureInformation.js
```diff
@@ -18,9 +18,21 @@ const FeatureInformation = ({ clickedFeatureInfo }) => {
   const { t } = useTranslation();
   const dispatch = useDispatch();
   const [featureIndex, setFeatureIndex] = useState(0);
-  const features = clickedFeatureInfo.map(l => l.features).flat();
+  const clickedFeatures = clickedFeatureInfo.map(l => l.features).flat();
+
+  // Extract cluster
```
gevos: @oterral These are changes in core. Let's discuss if you have a better idea to show cluster popups (`with f.get('features')`).

mhaertwig: I added a comment to the ticket. For me, the flattening of custom features should not happen in here.

src/layers/ConstructionLayer/ConstructionLayer.js
```diff
@@ -0,0 +1,87 @@
...
+      }),
+    );
+  }
+
+  terminate() {
+    super.terminate();
+
+    if (this.visibilityKeys.length) {
+      this.visibilityKeys.forEach(key => unByKey(key));
+    }
```
oterral:
```
unByKey(this.visibilityKeys)
```

src/layers/ConstructionLayer/ConstructionLayer.js
```diff
@@ -0,0 +1,87 @@
...
+  isFeatureVisible(feature) {
+    const childLayer = this.children.find(
+      child =>
+        child.properties &&
+        child.properties.construction &&
+        child.properties.construction.art === feature.get('art') &&
+        child.properties.construction.ort === feature.get('ort'),
+    );
+
+    return childLayer && childLayer.getVisible() ? feature.getGeometry() : null;
```
oterral:
```suggestion
    return !!(childLayer && childLayer.getVisible());
```

src/config/topics.js
```diff
@@ -9,6 +10,10 @@ import defaultBaseLayers, {
   netzkarteShowcases,
   parks,
   stuetzpunktbahnhoefe,
+  constrUnterhalt,
+  constrAusbau,
+  constrSingleLayer,
+  constrClusterLayer,
```
oterral: I'm wondering if it's possible to only add one layer:

```
aConstructionLayer = {
     olLayer: ol.layer.Group({
         layers: [  ol.layer.Cluster, ol.layer.vector ]  //  the 2 layers are based on the vector source
     }),
     children: [
          constrUnterhalt,
          constrAusbau,
     ],
     properties: {
        hideInlegend:true
     }
}
```

What do you think ?  I don't know if it's better or not.

src/layers/ConstructionLayer/ConstructionClusterLayer.js
```diff
@@ -0,0 +1,80 @@
...
+    const features = format.readFeatures(data);
+    this.olLayer
+      .getSource()
+      .getSource()
+      .clear();
+    this.olLayer
+      .getSource()
+      .getSource()
+      .addFeatures(features);
+  }
```
oterral: Shouldn't the addFeatures function goes in the `ConstructionLayer` ?  since we use it in the smae source.

src/components/FeatureInformation/FeatureInformation.js
```diff
@@ -18,9 +18,21 @@ const FeatureInformation = ({ clickedFeatureInfo }) => {
...
+  const features = clickedFeatures
+    .map(f => (f.get('features') ? f.get('features') : f))
```
oterral:
```suggestion
    .map(f => (f.get('features') || f))
```

src/components/FeatureInformation/FeatureInformation.js
```diff
@@ -18,9 +18,21 @@ const FeatureInformation = ({ clickedFeatureInfo }) => {
...
+    .flat();
   const feature = features[featureIndex];
-  const info = clickedFeatureInfo.find(i => i.features.includes(feature));
+
+  const info = clickedFeatureInfo.find(i => {
+    // Extract cluster
+    return i.features
+      .map(f => (f.get('features') ? f.get('features') : f))
```
oterral:
```suggestion
      .map(f => (f.get('features') || f))
```

src/layers/ConstructionLayer/ConstructionLayer.js
gevos: For the cluster layer, this is the [geometryFunction](https://openlayers.org/en/latest/apidoc/module-ol_source_Cluster-Cluster.html) and must return a geometry

```javascript
new ClusterSource({
    distance: 90,
    source: this.olLayer.getSource(),
    geometryFunction: this.isFeatureVisible,
 }),
```
I will rename the function.

src/layers/ConstructionLayer/ConstructionClusterLayer.js
gevos: These two different sources:
`this.olLayer.getSource()` for the single layer and `this.olLayer.getSource().getSource()` for the cluster layer as the cluster source is added as source to the vector layer:

```javascript
this.olLayer.setSource(
  new ClusterSource({
    distance: 90,
    source: this.olLayer.getSource(),
    geometryFunction: this.geometryFunction,
  }),
);
```
I will simplify it.

src/components/FeatureInformation/FeatureInformation.js
gevos: I suggest to flatten the features in `getFeatureInfoAtCoordinate()` of the `ConstructionLayer`, However, see question there...

src/layers/ConstructionLayer/ConstructionLayer.js
```diff
@@ -0,0 +1,202 @@
...
+        .map(f => (f.get('features') ? f.get('features') : f))
+        .flat();
+
+      return Promise.resolve({
+        features,
+        layer: data.layer,
+        coordinate: data.features.length
+          ? data.features[0].getGeometry().getCoordinates()
+          : data.coordinate,
+      });
```
gevos: In the Popup component, this coordinate isn't used as all features are Points (see [line](https://github.com/geops/trafimage-maps/blob/master/src/components/Popup/Popup.js#L28)). For clusters, this results in strange positions of the popup (if we use this approach with `getFeatureInfoAtCoordinate()`). @oterral Would it be possible always use a given coordinate in a feature info?

src/components/FeatureInformation/FeatureInformation.js
gevos: Removed

src/components/FeatureInformation/FeatureInformation.js
gevos: Removed

src/config/topics.js
gevos: I will check this.

gevos: Is your goal to save lines in layers.js?

I don't understand exactly. We still need to configure the `minResolution` and `maxResolution` as well as the GeoJSON loading for both layers. And `ol.layer.Cluster` does not exist. What exactly do you mean?

In order to save lines, we could
* change the `Layer` in `react-spatial` to accept `children` when initializing, see [line](https://github.com/geops/react-spatial/blob/master/src/layers/Layer.js#L37)
* add the "group" layers to the `ConstructionLayers`

```javascript
export const constrSingleLayer = new ConstructionLayer({
  name: 'ch.sbb.construction.single',
  key: 'ch.sbb.construction.single',
  visible: true,
  properties: {
    hideInLegend: true,
    popupComponent: 'ConstructionPopup',
  },
  children: [
    constrUnterhalt,
    constrAusbau,
  ],
  maxResolution: 305.748113141,
});
```

(and the same for constrClusterLayer)

oterral: yes ol.layer.Cluster doesn't exist. it was a shortcut to an ol layer with a cluster source.
I just meant, can we use a common vector source for both layers ( cluster and the vector) ?  so the data will be loaded only once.

oterral: The goal is to create one react-spatial laye instead of 2 but maybe it's not possible.

oterral: and yes react-sptial should be able to manage children from the constructor.

src/layers/ConstructionLayer/ConstructionLayer.js
oterral: Can't you get the cluster info ?

gevos: @oterral I would like, but it is always overridden by the geometry coordinate of the first feature, see [line](https://github.com/geops/trafimage-maps/blob/master/src/components/Popup/Popup.js#L28)

src/config/topics.js
gevos: > and yes react-sptial should be able to manage children from the constructor.

@oterral this would require changes in react-spatial: [PR](https://github.com/geops/react-spatial/pull/339)

src/layers/ConstructionLayer/ConstructionLayer.js
oterral: I guess you can change that in Popup and take the centroid of all features.

src/components/FeatureInformation/FeatureInformation.scss
```diff
@@ -5,12 +5,12 @@
...
     background-color: #f5f5f5;
     font-weight: bold;
-    height: 40px;
+    min-height: 40px;
 
     span {
       display: flex;
       align-items: center;
-      padding-left: 10px;
+      padding: 4px 0 4px 10px;
```
gevos: I would prefer these core two changes as they adjust the title height for all topics, if necessary.

src/components/Popup/Popup.js
```diff
@@ -25,7 +25,10 @@ const Popup = () => {
 
   const { coordinate, features } = clickedFeatureInfo[0];
   const geom = features[0].getGeometry();
-  const coord = geom instanceof Point ? geom.getCoordinates() : coordinate;
+  const coord =
+    features.length === 1 && geom instanceof Point
+      ? geom.getCoordinates()
+      : coordinate;
   const mapRect = map.getTarget().getBoundingClientRect();
```
gevos: I've also changed the core popup:
* This shows the popup at the click position when multiple features are clicked. Previously, the coordinate of the first feature was used, although the popup position does not update when going to the next feature in the popup
* And this allows to set a custom coordinate for cluster features.

src/config/layers.js
```diff
@@ -482,6 +483,167 @@ netzkarteShowcases.setChildren([
...
+  key: 'ch.sbb.construction.cluster',
+  visible: true,
+  properties: {
+    hideInLegend: true,
+    popupComponent: 'ConstructionPopup',
+    cluster: true,
+  },
+  children: [constrUnterhalt, constrAusbau],
+  minResolution: 305.748113141,
+});
```
oterral: So did you try 2 merge those 2 layers in one to load data only once ?

gevos: not yet

oterral: since it's not blocker , I'll merge this one , if you have time you can try it  in another PR.
 ok @gevos?

gevos: > since it's not blocker , I'll merge this one , if you have time you can try it in another PR.
ok @gevos?

This is a good idea and if @mhaertwig is okay with it.

mhaertwig: Yes, loading the data only once makes sense! @gevos please create a ticket for this.

PR #257, opened 2019-11-12
Remove ie11 unsupported 'outline-offset' and use a z-index instead fo…

edellucien:
…r barrier-free


oterral:
On small screen the first topic and the bg selector have a border missing on top

PR #262, opened 2019-11-13
Casa updates

mhaertwig:


src/layers/CasaLayer/CasaLayer.js
```diff
@@ -0,0 +1,172 @@
...
+    this.map.forEachFeatureAtPixel(pixel, f => {
+      featuresAtPixel.push(f);
+    });
+
+    const isTopFeature = this.olLayer
+      .getSource()
+      .getFeatures()
+      .includes(featuresAtPixel.slice(0, 1)[0]);
+
+    if (!isTopFeature) {
```
oterral: You could use `getFeaturesAtPixel`  with a` layerFIlter`, it will simplify:
https://openlayers.org/en/latest/apidoc/module-ol_Map-Map.html#getFeaturesAtPixel

PR #273, opened 2019-11-15
Gevos/layer info

gevos:


gevos:
I've added an "example" for a layer (Passagierfrequenzen) and for a topic (Netzkarte) to demonstrate the usage of the LayerInfo components. As soon as this PR is merged, I will add LayerInfo components for the remaining layers/topics with broken html tags.

PR #274, opened 2019-11-15
Add FilterButton & FollowButton from react-transit

edellucien:


edellucien:
Requires update of react-transit: https://github.com/geops/react-transit/pull/115

PR #281, opened 2019-11-18
Fix barrier free ie11 and smal screen

edellucien:


src/components/Collapsible/Collapsible.scss
```diff
@@ -1,5 +1,4 @@
 .wkp-collapsible-vertical {
-  overflow: hidden;
```
mhaertwig: @edellucien this breaks the style when collapsing / uncollapsing the menu.

src/components/Menu/MenuItem.scss
```diff
@@ -16,7 +16,6 @@
   border: 2px solid #666;
   background: white;
   max-height: 1000px;
-  overflow: hidden;
```
mhaertwig: @edellucien same here.

PR #304, opened 2019-11-22
Add lineKilometer search

edellucien:


edellucien:
Possibility to search for a lineKilometer with e.g:
'200 12'
additionaly to line search '200', and sub line search '200 12-35'

PR #306, opened 2019-11-22
Use plural or singular for results translations in search

edellucien:


src/components/Search/Search.js
```diff
@@ -78,7 +78,9 @@ function Search() {
                       {t(section)}
                     </div>
                     <div className="wkp-search-section-count">
-                      {t('overall_result', { count })}
+                      {count > 1
+                        ? t('overall_results', { count })
+                        : t('overall_result', { count })}
```
oterral: doc first : https://www.i18next.com/translation-function/plurals

oterral: Use `overall_result` and ` overall_result_plural`

edellucien: I didnt't know this functionality! It is great :)

PR #316, opened 2019-11-25
Re-order handicap infos in handicap popup

edellucien:


edellucien:
This character is actually wrong in cartaro, probably in the import.

PR #317, opened 2019-11-25
Delete public/img directory as absolute URLs are used

gevos:


gevos:
WIP: I have to make sure that the images are still visible locally.

src/layerInfos/BehigTopicInfo/BehigTopicInfo.js
```diff
@@ -12,7 +12,7 @@ const defaultProps = {};
 const BehigTopicInfo = ({ language, t }) => {
   const img = (
     <img
-      src={`/img/topics/behig/behig_legend_${language}.jpg`}
+      src={`${process.env.REACT_APP_STATIC_FILES_URL}/img/topics/behig/behig_legend_${language}.jpg`}
       draggable="false"
```
gevos: @oterral I have directly used the URL from process environment, such as in the DeparturePopup (see [line](in https://github.com/geops/trafimage-maps/blob/master/src/popups/DeparturePopup/DeparturePopupContent.js#L138]). I'm not sure if this is a good idea.

gevos:
> WIP: I have to make sure that the images are still visible locally.

Done.

PR #319, opened 2019-11-27
Construction: Use only one source to only have one request

gevos:


src/layers/ConstructionLayer/ConstructionLayer.js
```diff
@@ -19,65 +20,70 @@ import { unByKey } from 'ol/Observable';
...
-      minResolution: options.minResolution,
-      maxResolution: options.maxResolution,
+    const singleLayer = new OLVectorLayer({
+      style: (f, r) => this.styleSingle(f, r),
+      maxResolution: 305.748113141,
+    });
+
+    const clusterLayer = new OLVectorLayer({
+      style: (f, r) => this.styleCluster(f, r),
+      minResolution: 305.748113141,
```
oterral: It makes sens to have one variable for the 2 305.748113141 values

gevos: Done

PR #318, opened 2019-11-26
WIP Fix toggling mapboxlayer

edellucien:

PR #320, opened 2019-11-27
Fix visibility of mapbox layer

oterral:
Replace https://github.com/geops/trafimage-maps/pull/318


edellucien:
Great solution thanks! I tried to break the layer toggle unsuccessfully :+1:

PR #325, opened 2019-11-28
WIP Gevos/infofpw

gevos:
Use https://github.com/geops/trafimage-maps/pull/458 for test


src/config/topics.js
```diff
@@ -142,6 +143,18 @@ export const showcases = {
   layerInfoComponent: 'ShowcasesTopicInfo',
 };
 
+export const infofpw = {
+  name: 'ch.sbb.infofpw',
```
gevos: Currently, this topic is still visible, but it must be private.

src/popups/InfoFPWPopup/InfoFPWPopup.js
```diff
@@ -0,0 +1,114 @@
...
+    const { imageIndex } = this.state;
+
+    let text;
+    let image;
+    let pagination;
+
+    const pureText = feature.get(`${language}_text`) || feature.get('de_text');
+    if (pureText) {
+      // eslint-disable-next-line react/no-danger
+      text = <div dangerouslySetInnerHTML={{ __html: pureText }} />;
```
gevos: As this will be a private topic, we could dare to use it. What do you think?
Currently, the customer can generate HTML code in Cartaro with <strong>, <br />, <ul>, <ol>, <li>, and different kind of <a>. On production, all of these are used.

.env
```diff
@@ -1,6 +1,8 @@
 REACT_APP_PWA=inactive
 REACT_APP_CARTARO_URL=//cartaro2.stag.trafimage.ch/api/v1/
+REACT_APP_CARTARO_1_URL=//cartaro.trafimage.ch
```
gevos: I had to add a second Cartaro url to the old Cartaro.

src/popups/InfoFPWPopup/InfoFPWPopup.js
```diff
@@ -0,0 +1,114 @@
...
+        <img
+          src={`${process.env.REACT_APP_CARTARO_1_URL}/${photoPaths[imageIndex]}`}
+          draggable="false"
+          alt={t('Kein Bildtext')}
+        />
+      );
+
+      if (photoPaths.length > 1) {
+        pagination = (
+          <div className="wkp-pagination-wrapper">
```
gevos: Currently, I'm using the same classes from the component `<FeatureInformation>`. So, we do not have to change styling at two places. However, I don't think that this is good code style. What about a pagination component?

src/popups/InfoFPWPopup/InfoFPWPopup.js
```diff
@@ -0,0 +1,114 @@
...
+      image = (
+        <img
+          src={`${process.env.REACT_APP_CARTARO_1_URL}/${photoPaths[imageIndex]}`}
+          draggable="false"
+          alt={t('Kein Bildtext')}
+        />
+      );
+
+      if (photoPaths.length > 1) {
+        pagination = (
```
gevos: With the new baselayer and the new resolutions, it is possible that a popup opens with more than one feature. And if a feature has images (currently only on stag), we would see two pagination lines with `1 of 2`. As a solution, we could change the data. We could restrict the popup to only one feature. Or we could change the styling. What do you think?

src/popups/InfoFPWPopup/InfoFPWPopup.js
gevos: @oterral What do you think?

oterral: Fine for me

src/layerInfos/InfoFPWTopicInfo/InfoFPWTopicInfo.js
```diff
@@ -0,0 +1,51 @@
...
+      <span>
+        L’infografica fornisce una panoramica delle modifiche e dello sviluppo
+        dell’offerta dal cambiamento d’orario 2019.
+      </span>
+    ),
+  };
+
+  return (
+    <div>
+      {desc[language] ? desc[language] : desc.de}
```
oterral:
```suggestion
      {desc[language] || desc.de}
```

src/popups/InfoFPWPopup/InfoFPWPopup.js
oterral: Yes using the same pagination is not so nice.

Ideally a simple slideshow should be nice for images: https://react-slideshow.herokuapp.com/.
The lighter you get the better it is.

oterral: You can also create your own component, or just change the style of what you've already done (maybe the fatetr way).

gevos: I have integrated the react-slideshow-image lib.

src/popups/InfoFPWPopup/InfoFPWPopup.js
gevos: > I have integrated the react-slideshow-image lib.

So, this is obsolete.

gevos:
> * adds a max-height to the popup content.
> 
> * change the style of the image pagination or create a slideshow component for images.
> 
> * can you fix the display of the collapse/expand arrow in the layer tree. When the topic has no visible layer in the legend it shouldn't be displayed thx

Done.

src/popups/InfoFPWPopup/InfoFPWPopup.scss
```diff
@@ -0,0 +1,22 @@
+.wkp-infofpw-popup {
+  div {
+    line-height: 1.42857143;
```
oterral: are you sure of this ?  I don't want somebody complains thatn the line-height are different between popup.

gevos: In the old WKP, the line-height was generally 1.4... However, we would have a more compact popup without this line-height. Especially, as we use a popup with a max-height now instead of a sidebar with a variable height compared to the old WKP.

I have removed the line-height for both, the infofpw popup and the parks popup (here: replaced by padding).

src/components/FeatureInformation/FeatureInformation.scss
```diff
@@ -25,7 +25,17 @@
   }
 
   .wkp-feature-information-body {
-    padding: 10px;
+    & > * {
```
gevos: I have changed/fixed the popup styling (core, construction, and infofpw), see commits 7e716dc09cae2c34fc18f2fde9f9a7b4c85511cf and feab723166c2b24ff5e4c4c6148d808d826c0d94

src/components/FeatureInformation/FeatureInformation.scss
```diff
@@ -25,7 +25,17 @@
...
+      padding: 10px;
+    }
+
+    & > *:first-child {
+      padding-bottom: 0;
+    }
+
+    & > *:last-child {
+      padding-bottom: 10px;
```
oterral: this one is not necessary since you already apply a padding to all child

src/components/FeatureInformation/FeatureInformation.scss
oterral: please use div instead of *

src/components/FeatureInformation/FeatureInformation.scss
```diff
@@ -25,7 +25,17 @@
   }
 
   .wkp-feature-information-body {
-    padding: 10px;
+    & > * {
+      padding: 10px;
+    }
+
+    & > *:first-child {
```
oterral:
```suggestion
    & > div:first-child {
```

src/components/FeatureInformation/FeatureInformation.scss
oterral: ah yes it's necessary if there is only one child

src/components/FeatureInformation/FeatureInformation.scss
```diff
@@ -25,7 +25,17 @@
...
+      padding-bottom: 0;
+    }
+
+    & > *:last-child {
```
oterral:
```suggestion
    & > div:last-child {
```

src/components/FeatureInformation/FeatureInformation.scss
oterral:
```suggestion
    & > div {
```

src/components/FeatureInformation/FeatureInformation.scss
```diff
@@ -25,7 +25,17 @@
   }
 
   .wkp-feature-information-body {
-    padding: 10px;
```
gevos: This is a core change.

src/components/TopicMenu/TopicMenu.js
```diff
@@ -192,7 +192,13 @@ class TopicMenu extends PureComponent {
             <div
               className={`wkp-layer-toggler ${collapsed ? 'collapsed' : ''}`}
               style={{
-                display: topic.key === activeTopic.key ? 'block' : 'none',
```
gevos: Another core change to remove the topic arrow if there aren't any layers listed.

src/config/topics.js
gevos: This will be solved in another ticket.

src/components/FeatureInformation/FeatureInformation.scss
```diff
@@ -25,7 +25,17 @@
...
+    & > div {
+      padding: 10px;
+    }
+
+    & > div:first-child {
```
mhaertwig: @gevos please make the style definitions more explicit. Styling `div`'s without classes makes extending and changing styles very difficult.

gevos: @mhaertwig I cannot add class names as I do not know them.

There are two cases:

**Without pagination**, the last child is the custom topic popup where the className is always different, e.g. `wkp-infofpw-popup` or another class in another topic:
```
& > div:last-child {
  padding-bottom: 10px;
}
```

**With pagination**, the last child is the pagination  `wkp-pagination-wrapper`:
```
& > div:first-child { 
  padding-bottom: 0;
}

& > div:last-child {
  padding-bottom: 10px;
}
```


I could write this instead, but the first general line would remain:

```
& > div:last-child,
& > div.wkp-pagination-wrapper:last-child {
  padding-bottom: 10px;
}
```

gevos: To solve this, I have added a className property to each popup.

gevos:
@mhaertwig Please write a comment if you would like to merge this ticket.
Then I will 

- deactivate this ticket in topics.js
- merge this PR
- activated this topic in the new branch https://github.com/geops/trafimage-maps/tree/internal-topics

@edellucien @oterral f.y.i.

gevos:
WIP: I will add the old cartaro url to TrafimageMaps and to redux which allows to set instance urls

oterral:
ch.sbb.infofpw topic seems already  implemented in our version on gitlab. Please reopen if it makes sense

PR #327, opened 2019-11-29
Use svg sbb logo instead.

edellucien:


src/components/Header/Header.scss
```diff
@@ -14,5 +14,10 @@
 
   .wkp-header-right {
     margin-right: 10px;
+
+    svg {
+      display: flex;
```
oterral: wkp-header-right should be display `flex` not the svg.

src/components/Header/Header.scss
```diff
@@ -14,5 +14,10 @@
 
   .wkp-header-right {
     margin-right: 10px;
+
```
oterral:
```suggestion
      display: flex;
```

src/components/Header/Header.scss
oterral:
```suggestion
      display: flex;
```

PR #328, opened 2019-11-29
Add link to Developer portal in footer.

edellucien:


src/components/Footer/Footer.js
```diff
@@ -44,6 +44,13 @@ const Footer = () => {
         <ActionLink onClick={() => dispatch(setDialogVisible('Impressum'))}>
           {t('Impressum')}
         </ActionLink>
+        <a
+          href="https://doc.trafimage.ch"
+          rel="noopener noreferrer"
+          target="_blank"
+        >
+          {t('Developer Portal')}
```
oterral: Is there already translations?

src/components/Footer/Footer.scss
```diff
@@ -24,6 +24,12 @@
     margin: 0 5px;
   }
 
+  .wkp-footer-left {
+    a {
+      white-space: nowrap;
```
oterral: Why that ?

edellucien: TO avoid having a button two lined content e.g.:
'Developer
 Portal'

src/components/Footer/Footer.js
edellucien: Not yet, it will probably stay so on every language, I remove the 't'

src/components/Footer/Footer.scss
oterral: if I remove it nothing happens

oterral: ah yes it happens :)

PR #334, opened 2019-12-03
Gevos/construction link

gevos:


src/components/Link/Link.js
```diff
@@ -0,0 +1,35 @@
...
+const propTypes = {
+  href: PropTypes.string.isRequired,
+  children: PropTypes.node,
+};
+
+const defaultProps = {
+  children: null,
+};
+
+const Link = ({ href, children }) => {
```
gevos: I've added this new component on purpose as we will have several topics with popup links.

src/components/Link/Link.js
```diff
@@ -0,0 +1,35 @@
...
+      <LinkIcon />
+      {children}
+    </a>
+  );
+};
+
+Link.propTypes = propTypes;
+Link.defaultProps = defaultProps;
+
+export default compose(withTranslation())(Link);
```
oterral: You never use the `t` function so you don't need withTranslation.

In fucntionnal component you can use the useTranslation hook instead, it avoids the use of compose then you can return a memoized Component : export default React.memo(Link);

gevos: I have completely removed `compose(withTranslation())` as we do not use translations in this component.

PR #335, opened 2019-12-03
Rename topics

gevos:


src/lang/fr.json
```diff
@@ -76,7 +76,7 @@
   "ch.sbb.handicap": "Personen mit eingeschränkter Mobilität",
   "ch.sbb.infofpw": "Informations changement d'horaire",
   "ch.sbb.infofpw-desc": "L'infographie illustre les défis de chaque gare pour assurer l'exploitation. L'infographie pilote du changement d'horaire de décembre 2018 se limite aux régions de Zurich et de la Suisse orientale.<p>Responsable: P-V-STA-BWP, <a href=\"mailto:xp487@sbb.ch\">xp487@sbb.ch</a></p>",
-  "ch.sbb.infrastruktur": "Infrastructure ferroviaire suisse",
+  "ch.sbb.infrastruktur": "Carte de réseau: infrastructure ferroviaire",
```
oterral:
```suggestion
  "ch.sbb.infrastruktur": "Carte du réseau des infrastructures ferroviaires",
```
@edellucien  can you confirm ?

edellucien: Yes it is better

src/lang/fr.json
```diff
@@ -90,7 +90,7 @@
   "ch.sbb.lar-desc": "TINA - Système d’information topographique réseau et offre Thème:<br/>Vue d’ensemble de l’offre et du réseau de l’Infrastructure incluant les différents horizons. Représentation des données prévisionnelles de l’offre (nombre de trains, matériel roulant, modifications d’un horizon à l’autre) et des projets majeurs sous forme de cartes topographiques/schématiques et de synopsis. ",
   "ch.sbb.mobz": "Mobilitätshubs",
   "ch.sbb.mobz-desc": "Mobilitätshubs",
-  "ch.sbb.netzkarte": "Carte du réseau des TP suisses",
+  "ch.sbb.netzkarte": "Carte de réseau: trafic voyageurs",
```
gevos:
```suggestion
  "ch.sbb.netzkarte": "Carte du réseau: trafic voyageurs",
```
@oterral What about this?

gevos: Lucien said that there is only a minimal difference in meaning between "de" und "du". So, I will leave it the way the translation service provided it.

src/lang/fr.json
gevos: Lucien said that there is only a minimal difference in meaning between "de" und "du". So, I will leave it the way the translation service provided it.

PR #338, opened 2019-12-04
REturn to netzkarte Popup on departure popup close with close button.

edellucien:
Allow to redirect to netzkarte popup, when closing departure popup.

### DIFF
--- a/src/components/FeatureInformation/FeatureInformation.js
+++ b/src/components/FeatureInformation/FeatureInformation.js
@@ -82,6 +82,9 @@ const FeatureInformation = ({ clickedFeatureInfo }) => {
             className="wkp-close-bt"
             onClick={() => {
               dispatch(setClickedFeatureInfo());
+              if (PopupComponent.actionOnClose) {
+                PopupComponent.actionOnClose();
+              }
             }}
           >
             <MdClose focusable={false} />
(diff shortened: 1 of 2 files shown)


src/components/FeatureInformation/FeatureInformation.js
```diff
@@ -82,6 +82,9 @@ const FeatureInformation = ({ clickedFeatureInfo }) => {
             className="wkp-close-bt"
             onClick={() => {
               dispatch(setClickedFeatureInfo());
+              if (PopupComponent.actionOnClose) {
+                PopupComponent.actionOnClose();
```
oterral: DidUnmount  is not called on DeparturePopup when we close the popup with the close button ?

edellucien: 'componentWillUnmount' is called when clicking on the close button, but as well on clicking outside of the popup.
This solution allows us to return to the netzkarte Popup only on click on close button.

src/popups/DeparturePopup/DeparturePopup.js
```diff
@@ -24,5 +48,9 @@ DeparturePopup.defaultProps = defaultProps;
 
 const composed = compose(withTranslation())(DeparturePopup);
 composed.renderTitle = feat => feat.get('name');
+// Trigerred on popup close with close button only.
+composed.actionOnClose = () => {
```
oterral: The namimg is a bit confusing since it applies only when we click on the close button.

edellucien:
Because the onCloseClick is called when we click on the close button in the popup's Header.
However here, we are hidding the header in the popup (because we don't want the default one), and adding the header in the FeatureInformation component.

src/popups/DeparturePopup/DeparturePopup.js
edellucien: You are right, I renamed it 'onCloseBtClick'.

PR #358, opened 2019-12-11
Add lang param to the permalink.

edellucien:


src/components/Permalink/Permalink.js
```diff
@@ -181,6 +202,13 @@ class Permalink extends PureComponent {
     });
   }
```
mhaertwig: A function to call a function that is only used once? Keeping it simple would reduce code and increase readability.

edellucien: The function is made because calling setState directly in componentDidUpdate is restricted by eslint, even though called within a if.

PR #371, opened 2019-12-13
Lucien/barrierfree enhancement

edellucien:


edellucien:
Requires https://github.com/geops/react-spatial/pull/402 to be merged

src/components/Popup/Popup.js
```diff
@@ -31,7 +31,7 @@ const Popup = () => {
 
   return (
     <RSPopup
-      showHeader={false}
+      renderHeader={() => {}}
```
friedjoff: Maybe we should set a default prop for this in react-spatial to avoid this?

src/components/Dialog/Dialog.js
```diff
@@ -26,15 +26,28 @@ function Dialog(props) {
...
+  const dialogRef = useRef(null);
+  let activeElement;
+
+  const escFunction = e => e.which === 27 && dispatch(setDialogVisible());
 
   const registerEsc = isRegister => {
     if (isRegister) {
       document.addEventListener('keydown', escFunction, false);
+
+      activeElement = document.activeElement;
```
friedjoff: Is it possible to add a test for this?

src/components/Popup/Popup.js
edellucien: Then by default the Header will not be displayed

src/components/Share/Share.js
```diff
@@ -51,7 +51,7 @@ const Share = () => {
     <div className="wkp-share">
       <SharePermalinkButton />
       {renderConf(config[0], t)}
-      <CanvasSaveButton title={t('Karte als Bild speichern')} map={map}>
+      <CanvasSaveButton aria-label={t('Karte als Bild speichern')} map={map}>
```
oterral: No more title ?

oterral: React doesn't complain about aria-label ? should not be ariaLabel ?

src/components/Dialog/Dialog.js
```diff
@@ -26,15 +26,28 @@ function Dialog(props) {
...
 
-  const escFunction = () => {
-    dispatch(setDialogVisible());
-  };
+  const dialogRef = useRef(null);
+  let activeElement;
+
+  const escFunction = e => e.which === 27 && dispatch(setDialogVisible());
 
   const registerEsc = isRegister => {
```
oterral: Since you use this function only once, I will move the content inside the useEffect directly.
This function will be overriden on each render, I think it's not good. Same for `activeElement` on each render the variable will be set to undefined so define it inside the useEffect. 

The code works because we mount/unmount the component and never rerender it .

src/components/Share/Share.js
edellucien: Nop It complains about ariaLabel (taken as a prop and not as the aria-label of the element) :)

edellucien: I add the title to the children (title prop has been removed from the component: https://github.com/geops/react-spatial/commit/f49dbfe9f9e9063c77d761fc9e96ac45a59144a9#diff-84392269f1dcee3ac19b57cd31741243)

src/components/Dialog/Dialog.js
edellucien: It seems rather hard to test and this is part of the release plan for tomorrow

edellucien:
Changed

src/components/Dialog/Dialog.js
```diff
@@ -26,19 +26,32 @@ function Dialog(props) {
...
+        if (dialogFocusables.length) {
+          // Focus the first focusable element in the popup.
+          dialogFocusables[0].focus();
+        }
+      } else {
+        document.removeEventListener('keydown', escFunction, false);
+        // Re focus the element that opened the dialog.
+        activeElement.focus();
+      }
+    };
```
oterral: I meant the content of the function not the function itself.
you don't need to encapsulate it into a function

edellucien: I see what you mean now sorry :)

src/components/TrafimageMaps/TrafimageMaps.scss
```diff
@@ -67,6 +67,14 @@
     left: 0;
   }
 
+  .rs-popup {
+    .rs-popup-container {
```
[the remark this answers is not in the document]
edellucien: Such rule could be interesting but we should integrate directly in our husky, and not in the merge request, to be able to spot it fast.

src/components/Dialog/Dialog.js
edellucien: @friedjoff Would it make sense to rather test this with cypress?

edellucien:
Another commit after mario's commit:
 - alternative text for close icon
 - aria-expanded on other togglers

src/components/Dialog/Dialog.js
friedjoff: Ok, let's use cypress :)

PR #377, opened 2019-12-16
State in handi popup when a station isn't without stairs.

edellucien:


src/lang/fr.json
```diff
@@ -250,6 +250,7 @@
   "Naturerlebnispark": "Parc naturel périurbain",
   "Nein": "Non",
   "Nicht barrierefreier Bahnhof": "Gare non accessible",
+  "nicht treppenfrei": "Pas sans escalier",
```
oterral: @edellucien is that fench ? :)

edellucien: Litteral translation of 'nicht treppenfrei' or 'not without stairs' indeed.
But even in other language it seems weird.

src/popups/HandicapPopup/HandicapPopup.js
```diff
@@ -32,6 +31,11 @@ function HandicapPopup({ feature }) {
 
   // build string for equipment
   const equipment = [];
+
+  equipment.push(
```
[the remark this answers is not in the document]
edellucien:
```suggestion
  equipment.unshift(
```

src/lang/fr.json
oterral: `Presence d'escaliers`  ou `avec escaliers` ou `Escaliers existant`

edellucien: tu manies la prose comme personne :smile:

PR #385, opened 2019-12-18
Use apiKey by default

oterral:


oterral:
@mhaertwig PR updated

PR #390, opened 2019-12-19
Switch search to react-transit

friedjoff:
- [x] create new react-transit release and update trafimage-maps


friedjoff:
@lb1c is it possible to make hound less noisy? 99% of this stuff is already covered by ESLint and Prettier which runs in our GitHub action CI.

lb1c:
`hound` was just a test. but since it's more disturbing i'll deactivate it for now

friedjoff:
> `hound` was just a test. but since it's more disturbing i'll deactivate it for now

Thanks! In general I like the idea, but maybe we should use more specific apps like the ImgBot.

edellucien:
We should ensure this MR is a WIP, until we correct the react-transit Search component.
The search functionality isn't quiet prefect yet.

src/components/Search/Search.js
```diff
@@ -1,147 +1,103 @@
...
-    };
-  }, []);
+function Search({ apiKey }) {
+  const dispatch = useDispatch();
+  const { t } = useTranslation();
+  const { layerService, map, searchEngines } = useSelector(state => ({
+    layerService: state.app.layerService,
+    map: state.app.map,
+    searchEngines: state.app.activeTopic.searchEngines,
+  }));
```
oterral: this generates always a rerender of the component . I will separate in 3 useSelector instead.

oterral:
Is this still actual ? @mhaertwig ?

PR #409, opened 2020-01-09
Gevos/permalink layers

gevos:


edellucien:
LGTM, there is a conflict however

PR #416, opened 2020-01-10
Add 'mobz' & 'What-if mobz topics

edellucien:


src/popups/MobzPopup/MobzPopup.scss
```diff
@@ -0,0 +1,9 @@
+.wkp-mobz-popup {
+  height: 20px;
```
gevos: Is this height necessary?

src/layerInfos/MobzLayerInfo/MobzLayerInfo.js
```diff
@@ -0,0 +1,263 @@
...
+        </span>
+      ),
+    },
+  };
+
+  return (
+    <div className="wkp-mobz-layer-info">
+      <div>
+        {img}
+        {desc[category][language]}
```
gevos: As long as the description does not contain html tag, I would prefer to get the (translated) description from the translation files, such as in https://github.com/geops/trafimage-maps/blob/master/src/layerInfos/NetzkarteTopicInfo/NetzkarteTopicInfo.js#L15

src/popups/MobzPopup/MobzPopup.js
```diff
@@ -0,0 +1,31 @@
...
+const MobzPopup = ({ feature, t }) => {
+  const category = feature.get('kategorisierung');
+
+  return (
+    <div className="wkp-mobz-popup">
+      <img
+        alt={t(category)}
+        src={`${process.env.REACT_APP_STATIC_FILES_URL}/img/layers/mobz/${category}.png`}
+      />
+      <span>{t(category)}</span>
```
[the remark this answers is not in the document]
edellucien: I missed this property indeed! Thanks :)

PR #433, opened 2020-01-16
Remove apiKey from .env

edellucien:


edellucien:
> API_KEY variable still use in src/WebComponent.js

Indeed, I delete it, I thought I should leave it to be used in gitlab.

oterral:
Maybe we need it I don't know, test to be sure

edellucien:
It is ok, gitlab uses the config.js' api_key

PR #444, opened 2020-01-21
Display user name on login button when logged in

edellucien:


src/model/app/reducers.js
```diff
@@ -105,7 +105,10 @@ export default function app(state = getInitialState(), action) {
     case SET_PERMISSIONS:
       return {
         ...state,
-        permissions: [...action.data],
+        permissions: action.data.permissions
```
mhaertwig: Please use one object containing user and permissions so it can be easily extended later.

PR #450, opened 2020-01-22
Adapt lock style for small screens

edellucien:


edellucien:
Temporary remove cypress action, due to a upflow github action problem.

PR #453, opened 2020-01-22
[WIP] Gevos/zweitausbildung

gevos:
# How to

[Test application](https://deploy-preview-626--dev-trafimage-maps.netlify.com) (netlify from https://github.com/geops/trafimage-maps/pull/626) as the topic is hidden in the LayerTree, now

# Open issues:

- [x] hidden, but public topic
- [x] routes dropdown layer always expanded (see https://github.com/geops/react-spatial/pull/537)
- [x] add images to routes dropdown
- [x] reset routes dropdown on singleclick
- [x] routes features info: meterRad zoom dependent
- [x] pulse style for highlighting single features in pois cluster
- [x] popup title for pois when both pois layers are hit
- [x] increase popup width
- [x] popup scrollbar
- [x] dynamic popup height (similar to sidebar in old wkp)? - no, we use the popup
- [x] remove forceUpdate() in routes popup when state changes
- [x] pois animation circle sometimes too big (seen in IE11)
- [x] underlying default geometry at pois animation
- [x] https://github.com/geops/trafimage-maps/pull/648
- [x] https://github.com/geops/trafimage-maps/pull/647
- [x] https://github.com/geops/trafimage-maps/pull/651
- [x] alignment of layer info icons/labels

# Others

- [x] IE11 tested.
- [x] Tests added. - No, see https://github.com/geops/trafimage-maps/pull/453#discussion_r400831289


src/popups/ZweitausbildungRoutesPopup/ZweitausbildungRoutesPopup.js
```diff
@@ -0,0 +1,126 @@
...
+        'highlight',
+        highlightFeatures[i].get('bezeichnung') ===
+          singleFeature.get('bezeichnung'),
+      );
+      highlightFeatures[i].changed();
+    }
+
+    this.setState({ features: singleFeatures });
+
+    this.forceUpdate();
```
gevos: @oterral Do you know why the re-rendering only works with `this.forceUpdate()`?

src/layers/ZweitausbildungRoutesHighlightLayer/ZweitausbildungRoutesHighlightLayer.js
```diff
@@ -0,0 +1,244 @@
...
+      [parseInt(coordinate[0], 10), parseInt(coordinate[1], 10)],
+      'EPSG:3857',
+      'EPSG:21781',
+    );
+
+    return fetch(
+      `${this.geoServerUrl}?` +
+        'service=WFS&version=1.0.0&request=GetFeature&' +
+        `typeName=trafimage:${this.zweitProps.featureInfoLayer}&` +
+        // 'srsName=EPSG:3857&' +
```
edellucien: Can we remove this commented line?

src/popups/ZweitausbildungRoutesPopup/ZweitausbildungRoutesPopup.js
edellucien: The set state doesn;t set a new state but the same value?

edellucien: The setState doesn't set a new state but the same value?

src/lang/en.json
```diff
@@ -214,30 +240,33 @@
   "linienkilometer_plural": "Line kilometers",
   "liniensegment": "Line segment",
   "liniensegment_plural": "Line segments",
+  "Link Übersichtsliste Geografie im Produkte-eSpace VM HR-BIL": "Link to geographical overview list in the VM HR-BIL product eSpace",
   "login_url": "http://www.sbb.ch/en/meta/login.html",
   "Lokalisieren": "Localize",
+  "Länge": "Longitude",
+  "Länge": "Longitude",
```
edellucien: Longitude added twice

src/layers/ZweitausbildungRoutesHighlightLayer/ZweitausbildungRoutesHighlightLayer.js
gevos:
```suggestion
```

src/config/layers.js
```diff
@@ -795,6 +799,271 @@ export const grenzen = new Layer({
...
+            desc: 'ch.sbb.zweitausbildung.tourist.routes.group-desc',
+            legend: {
+              image: 'legend_tourist_strecken.png',
+            },
+          },
+          layer: 'zweitausbildung_tourist_strecken_grouped_qry',
+        },
+      },
+      children: [
+        new ZweitausbildungRoutesHighlightLayer({
```
gevos: I have added the dropdown field as a child layer with the `layer.renderItemContent()` function. This dropdown should always be visible. I thought about adding an `expandWithParent` property to the layer. So, the LayerTree (react-spatial) always expands its child layers if one child has this property. @oterral what do you think?

gevos: > I thought about adding an `expandWithParent` property to the layer. So, the LayerTree (react-spatial) always expands its child layers if one child has this property. 

I should better use an `expandChildren` property to the layer where its children should be expanded. Otherwise, it would be too complicated.

Nevertheless, what do you think about this react-spatial change, @oterral?

src/popups/ZweitausbildungRoutesPopup/ZweitausbildungRoutesPopup.js
oterral: create a new array should avoid the use of forceUpdate:
```
this.setState({ features: [...singleFeatures] });
```

src/config/layers.js
oterral: yes a  boolean isAlwaysExpanded  that removes the expand functionnality and always display the children is a good idea.

oterral: @gevos

gevos:
>     * the select box breaks the menu css, define a fixed width should fix the issue

I have replaced the select box by a new `IconList` component which allows to use images

>     * the cursor pointer inside Touristische Reiseziele popup is not necessary, because nothing happens when we click.
The single features in the cluster are highlighted when you hover over the popup items.

src/config/layers.js
gevos: I will add the property which initially expands the children. 
@oterral Would you still show the expand arrow and offer the expand functionality?

gevos: @oterral https://github.com/geops/react-spatial/pull/537

oterral:
> > ```
> > * the select box breaks the menu css, define a fixed width should fix the issue
> > ```
> 
> I have replaced the select box by a new `IconList` component which allows to use images
> 
> > ```
> > * the cursor pointer inside Touristische Reiseziele popup is not necessary, because nothing happens when we click.
> > ```
> 
> The single features in the cluster are highlighted when you hover over the popup items.

Yes but cursor: pointer means you can click on it to make something happens. So just use the default cursor.

gevos:
> Yes but cursor: pointer means you can click on it to make something happens. So just use the default cursor.

Fixed

src/layers/ZweitausbildungPoisLayer/ZweitausbildungPoisLayer.js
```diff
@@ -0,0 +1,204 @@
...
+  animate(event) {
+    const duration = 200;
+    const maxRadius = 14;
+    const color = this.highlightFeature.get('color') || 'rgba(50, 50, 50, 0.8)';
+
+    const elapsed = event.frameState.time - this.animationStartTime;
+
+    if (elapsed < 0) {
+      return;
+    }
```
gevos: @oterral Do you have an idea why this is possible?

oterral: no idea

src/components/TopicLoader/TopicLoader.js
```diff
@@ -211,17 +218,25 @@ class TopicLoader extends Component {
...
+      if (flatLayers[i].setStyleConfig) {
         flatLayers[i].setStyleConfig(vectorTilesUrl, vectorTilesKey);
-      } else if (flatLayers[i].setCartaroUrl) {
+      }
+      if (flatLayers[i].setCartaroUrl) {
         flatLayers[i].setCartaroUrl(cartaroUrl);
       }
+
+      if (flatLayers[i].setLanguage) {
+        flatLayers[i].setLanguage(language);
```
gevos: Core change

src/components/TopicLoader/TopicLoader.js
```diff
@@ -211,17 +218,25 @@ class TopicLoader extends Component {
     for (let i = 0; i < flatLayers.length; i += 1) {
       if (flatLayers[i].setGeoServerUrl) {
         flatLayers[i].setGeoServerUrl(`${appBaseUrl}/geoserver/trafimage/ows`);
-      } else if (flatLayers[i].setGeoServerWMSUrl) {
+      }
```
gevos: Core changes

src/components/TopicMenu/TopicMenu.js
```diff
@@ -177,7 +177,9 @@ class TopicMenu extends PureComponent {
             }}
             renderItemContent={(layer, layerTreeComp) => (
               <>
-                {layerTreeComp.renderItemContent(layer)}
+                {layer.renderItemContent
+                  ? layer.renderItemContent(layerTreeComp)
+                  : layerTreeComp.renderItemContent(layer)}
```
gevos: Core change to use renderItemContent() from layer if available

src/components/FeatureInformation/FeatureInformation.js
```diff
@@ -78,7 +78,7 @@ const FeatureInformation = ({ featureInfo }) => {
         <div className="wkp-feature-information-header">
           <span>
             {PopupComponent && PopupComponent.renderTitle && feature
-              ? PopupComponent.renderTitle(feature)
+              ? PopupComponent.renderTitle(feature, t)
```
gevos: Core change

src/components/IconList/IconList.js
```diff
@@ -0,0 +1,125 @@
...
+  onSelect: PropTypes.func.isRequired,
+  t: PropTypes.func.isRequired,
+};
+
+const defaultProps = {
+  disabled: false,
+  selected: null,
+};
+
+class IconList extends PureComponent {
```
gevos: Core change: new IconList component to get a dropdown with images

src/WebComponent.js
```diff
@@ -161,6 +161,11 @@ const WebComponent = props => {
       tps.forEach(topic => {
         // eslint-disable-next-line no-param-reassign
         topic.active = topic.key === activeTopicKey;
+        // eslint-disable-next-line no-param-reassign
+        topic.hideInLayerTree =
+          topic.hideInLayerTree && topic.key === activeTopicKey
+            ? false
+            : topic.hideInLayerTree;
```
gevos: Core change

src/components/TopicLoader/TopicLoader.js
```diff
@@ -124,7 +130,10 @@ class TopicLoader extends Component {
       return;
     }
     const visibleTopics = topics.filter(
-      t => !t.permission || permissionsInfos.permissions.includes(t.permission),
+      t =>
+        (!t.permission ||
+          permissionsInfos.permissions.includes(t.permission)) &&
+        !t.hideInLayerTree,
```
gevos: Core change

src/components/IconList/IconList.js
```diff
@@ -0,0 +1,125 @@
...
+    });
+  }
+
+  renderOption(option) {
+    const { icons, t, options } = this.props;
+
+    const opt = option || (options.length ? options[0] : null);
+
+    const label = (
+      <div className="wkp-zweitausbildung-routes-dropdown-label">{t(opt)}</div>
```
gevos: I will change this (and the following lines) as zweitausbildung should not be part of `IconList`

src/components/IconList/IconList.js
```diff
@@ -0,0 +1,174 @@
...
+
+    if (displayOnTop) {
+      const rect = this.ref.current.getBoundingClientRect();
+      const listStyles = {
+        top: `${rect.y + rect.height}px`,
+        left: `${rect.x}px`,
+      };
+
+      const targetElements = document.getElementsByClassName(
+        'tm-trafimage-maps',
```
gevos: @oterral I'm not sure if this a good idea to reference this element. I would like an element which is in our scope but similar to body. So I can add this dropdown list to the very top of all elements.

However, this class name is also used in other components (css).

src/components/IconList/IconList.js
```diff
@@ -0,0 +1,174 @@
...
+        <div className="wkp-icon-list-modal">
+          <div className="wkp-icon-list-modal-pos" style={listStyles}>
+            {list}
+          </div>
+        </div>,
+        targetElements[0],
+      );
+    }
+
+    return <div className="wkp-icon-list-non-modal">{list}</div>;
```
gevos: Although, we don't need it, I haven't removed the previous version, where the dropdown is displayed inline (displayOnTop=false). I think, the inline version could be the default one. @oterral what do you think?

src/components/IconList/IconList.js
```diff
@@ -0,0 +1,174 @@
...
+          onSelect(option);
+          this.setState({
+            iconListVis: false,
+            selectedOption: option,
+          });
+        }}
+      />
+    );
+
+    if (displayOnTop) {
```
gevos: @oterral I thought about adding tests. But, they would only make sense if you could make a snapshot test after clicking on the dropdown button. But I haven't found a solution, yet. @oterral any idea?

src/components/IconList/IconList.js
oterral: @gevos does it matter if it's in our scope ?  document.body is fine I guess. Because it will be easier to get/set a position form/to the body. if you need to have the position depending on trafimage-maps elt , i will be more complicated. Try with document.body for now. We will after if we need more.

gevos: @oterral What about the WebComponent? I thought, it should be possible to integrate a trafimage-maps application into another page without using an IFrame. Adding an element to the body could have side-effects.

gevos: @oterral And there is another thing: I cannot use the styles from IconList.scss as they are in a scope of the class s466903671.

src/components/IconList/IconList.js
oterral: I guess you can keep it.

src/components/IconList/IconList.js
```diff
@@ -0,0 +1,174 @@
...
+        'tm-trafimage-maps',
+      );
+
+      if (!targetElements || !targetElements.length) {
+        return null;
+      }
+
+      return ReactDOM.createPortal(
+        <div className="wkp-icon-list-modal">
+          <div className="wkp-icon-list-modal-pos" style={listStyles}>
```
gevos, line 128: Setting the style does not work in IE11

gevos: Fixed

src/layers/ZweitausbildungPoisLayer/ZweitausbildungPoisLayer.js
```diff
@@ -0,0 +1,204 @@
...
+
+    this.olLayerHighlight = new OLVectorLayer({
+      map,
+      source: new OLVectorSource(),
+    });
+
+    this.olLayerHighlight.getSource().on('addfeature', () => {
+      this.animationStartTime = new Date().getTime();
+      unByKey(this.postrenderKey);
+      this.postrenderKey = this.olLayerHighlight.on('postrender', this.animate);
```
gevos, line 88: Unfortunately, there is the default style (small blue circle) below the grey animation circle. And I haven't found out how I can avoid this. When I set a style function for the olLayerHight or directly to the highlightFeature, the postrender callback isn't called. @oterral any ideas?

src/layers/ZweitausbildungPoisLayer/ZweitausbildungPoisLayer.js
```diff
@@ -0,0 +1,204 @@
...
+
+    const elapsed = event.frameState.time - this.animationStartTime;
+
+    if (elapsed < 0) {
+      return;
+    }
+
+    const elapsedRatio = elapsed / duration;
+    // radius will be 1 at start and 14 at end
+    const radius = easeOut(elapsedRatio) * maxRadius + 1;
```
gevos, line 109: Testing on IE11, the radius was sometimes bigger. I have to find out why.

gevos: Fixed.

src/layers/ZweitausbildungPoisLayer/ZweitausbildungPoisLayer.js
gevos: Fixed

src/components/IconList/IconList.js
oterral: Yes,  because of the styles is not possible I guess.

oterral:
on the test link the select box is a modal, is that normal? it feels really weird.

src/components/IconList/IconList.js
```diff
@@ -0,0 +1,174 @@
...
+      const targetElements = document.getElementsByClassName(
+        'tm-trafimage-maps',
+      );
+
+      if (!targetElements || !targetElements.length) {
+        return null;
+      }
+
+      return ReactDOM.createPortal(
+        <div className="wkp-icon-list-modal">
```
gevos, line 127: > on the test link the select box is a modal, is that normal? it feels really weird.

> This modal is weird.

@oterral yes, it is a modal. By using a modal, I wanted to avoid to update the opened dropdown position when the user scrolls the layer tree.

oterral: but we also can't  go out of this modal, unless we click on the select box. It's annoying. Can't you make the modal bg transparent and on a user interaction on it, you close the select box ?

gevos: see https://github.com/geops/trafimage-maps/pull/638/files#r403911688

src/layerInfos/ZweitausbildungLayerInfo/ZweitausbildungLayerInfo.js
```diff
@@ -0,0 +1,59 @@
...
+  const { infos } = properties.get('zweitausbildung');
+  const { title, legend } = infos;
+
+  return (
+    <div className="wkp-zweitausbildung-layer-info">
+      <div>{t(title)}</div>
+      {legend ? (
+        <div className="wkp-zweitausbildung-layer-info-legend">
+          <img
+            src={`${process.env.REACT_APP_STATIC_FILES_URL}/img/layers/zweitausbildung/${legend.image}`}
```
oterral, line 21: I don't think the use of process.env.REACT_APP_STATIC_FILES_URL is a good idea. The urls should come from a property . are we doing this in another layers?

oterral: or from redux

gevos:
> icon and label in layer's inof popup are note vertically aligned

@oterral I'm not sure what exactly should be aligned. I have made the layer info icons of Stationen Basis and Aufbau symmetrical: https://gitlab.geops.de/geops/trafimage-maps/-/merge_requests/67

gevos:
> > icon and label in layer's inof popup are note vertically aligned
> 
> @oterral I'm not sure what exactly should be aligned. I have made the layer info icons of Stationen Basis and Aufbau symmetrical, see gitlab MR

Symmetrical icons deployt on all instances

oterral:
@gevos For the algnment I meant this: https://github.com/geops/trafimage-maps/pull/453/commits/776e804d15cb14cad25474c43bef872a8c21b107

PR #454, opened 2020-01-22
Fix getSources only when mbmap is defined

gevos:


edellucien:
Probably already resolved here: https://github.com/geops/trafimage-maps/pull/442/files

gevos:
> Probably already resolved here: https://github.com/geops/trafimage-maps/pull/442/files

@edellucien No, it's not. This is why I used a similar commit message.

PR #479, opened 2020-02-06
Netzkarte popup: Fix coordinate data type and projection for transformation

gevos:


src/popups/NetzkartePopup/NetzkartePopup.js
```diff
@@ -127,7 +127,10 @@ function NetzkartePopup({ feature }) {
 
   const coordinates = transformCoords(
     feature.get('longitude') && feature.get('latitude')
-      ? [feature.get('longitude'), feature.get('latitude')]
+      ? [
+          parseFloat(feature.get('longitude'), 10),
+          parseFloat(feature.get('latitude'), 10),
+        ]
       : feature.getGeometry().getCoordinates(),
     'EPSG:21781',
```
oterral: I think we don't need the transformation anymore.

gevos: @oterral We still need the transformation as we can change the projection in the footer.
I have changed the data projection. The coordinates are now in EPSG:4326 instead of EPSG: 21781.

BTW: Currently, the initial projection in the popup is EPSG:4326, but the projection in the footer is EPSG:21781. I will have a look at it.

PR #480, opened 2020-02-07
Mouse position: Use default projection from redux in mouse position

gevos:
WIP: As soon as the react-spatial [PR](https://github.com/geops/react-spatial/pull/482) is merged, I can create a new react-spatial version and integrate it here.

### DIFF
--- a/src/components/Footer/Footer.js
+++ b/src/components/Footer/Footer.js
@@ -88,7 +88,7 @@ const Footer = () => {
             dispatch(setProjection(proj));
           }}
           projections={projections}
-          value={projection}
+          projectionValue={projection}
         />
         <ScaleLine map={map} />
         <Select
--- a/src/components/Footer/Footer.js
+++ b/src/components/Footer/Footer.js
@@ -30,6 +30,30 @@ const Footer = () => {
   const layerService = useSelector(state => state.app.layerService);
   const map = useSelector(state => state.app.map);
   const language = useSelector(state => state.app.language);
+  const projection = useSelector(state => state.app.projection);
+
+  const projections = [
+    {
+      label: 'CH1093 / LV03',
+      value: 'EPSG:21781',
+      format: c => `${t('Koordinaten')}: ${numberFormat(c)}`,
+    },
+    {
+      label: 'CH1093+ / LV95',
+      value: 'EPSG:2056',
+      format: c => `${t('Koordinaten')}: ${numberFormat(c)}`,
+    },
+    {
+      label: 'Web Mercator',
+      value: 'EPSG:3857',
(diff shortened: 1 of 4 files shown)


src/components/Footer/Footer.js
```diff
@@ -63,29 +87,10 @@ const Footer = () => {
...
-              format: c => `${t('Koordinaten')}: ${numberFormat(c)}`,
-            },
-            {
-              label: 'WGS 84',
-              value: 'EPSG:4326',
-              format: c =>
-                `${t('Koordinaten')}: ${c[0].toFixed(5)},${c[1].toFixed(5)}`,
-            },
-          ]}
+          projections={projections}
```
friedjoff: How about something like (not tested)?
```suggestion
          projections={[
            projections.find(p => p.value === projection.value,
            ...projections.filter(p => p.value === projection.value)
          ]}
```
This would avoid adding a new prop to MousePosition.

src/components/Footer/Footer.js
```diff
@@ -63,29 +87,10 @@ const Footer = () => {
...
+          initialIndex={projections.findIndex(
```
friedjoff: If we really want this prop, I would call it something like `initialProjection` and use the projection value.

src/components/Footer/Footer.js
gevos: You meant:
```suggestion
          projections={[
            projections.find(p => p.value === projection.value),
            ...projections.filter(p => p.value !== projection.value),
          ]}
```
But this would change the order of the values in the drop down list. I didn't want to do this.

src/components/Footer/Footer.js
gevos: Done

src/components/Footer/Footer.js
oterral: I agree with @gevos

src/components/Footer/Footer.js
```diff
@@ -7,29 +7,44 @@ import Copyright from 'react-spatial/components/Copyright';
...
+      label: 'Web Mercator',
+      value: 'EPSG:3857',
+      format: c => `${t('Koordinaten')}: ${coordinateHelper.meterFormat(c)}`,
+    },
+    {
+      label: 'WGS 84',
+      value: 'EPSG:4326',
+      format: c => `${t('Koordinaten')}: ${coordinateHelper.wgs84Format(c)}`,
+    },
+  ];
```
oterral: I would  put the projections array outside the functionnal component  so we create it only once.

gevos: @oterral But we need the translation function

oterral: yes I hadn't seen

PR #487, opened 2020-02-13
Update react-spatial to '0.1.1-beta.22'

edellucien:


gevos:
> LGTM, but have have missed a property name change:
> 
> #498

Done

PR #490, opened 2020-02-14
Ensure to not always pass netzkarte as activeTopic first.

edellucien:


src/WebComponent.js
```diff
@@ -145,6 +145,11 @@ const WebComponent = props => {
       console.error('You must provide a list of topics');
       return [];
     }
+
+    const urlTopic = window.location.pathname.replace('/', '');
+    if (urlTopic !== '' && urlTopic !== activeTopicKey) {
```
oterral:
```suggestion
    if (urlTopic && urlTopic !== activeTopicKey) {
```

PR #491, opened 2020-02-14
Switch max zoom to 15

edellucien:


edellucien:
> LGTM
> 
> @edellucien For the whole map or only for the infrastrukturkarte?

Good question actually, I asked to PO because it is a bit unclear.

edellucien:
You were right, Po confirmed, only the 'infrastructure topic' maximum zoom needs to be set to 15.

1/ Need to be merged, published and used in this branch:
- react-spatial update to be able to update Map viewOptions: https://github.com/geops/react-spatial/pull/494

2/ Need fix from latest react-spatial to be merged as well:
https://github.com/geops/trafimage-maps/pull/487

src/config/topics.js
```diff
@@ -140,6 +140,7 @@ export const behig = {
 export const infrastruktur = {
   name: 'ch.sbb.infrastruktur',
   key: 'ch.sbb.infrastruktur',
+  maxZoom: 14,
```
gevos:
```suggestion
  maxZoom: 15,
```
14 or 15?

PR #502, opened 2020-02-18
CASA baselayers

edellucien:


edellucien:
As PO asked, css work has been removed from this MR.

PR #504, opened 2020-02-18
Fix menu item icons with display flex instead of position absolute

edellucien:


edellucien:
Thanks for the comment. I adapted the code.

src/components/TopicMenu/TopicMenu.scss
```diff
@@ -74,11 +73,10 @@
...
 
     .wkp-info-topic-bt {
-      position: absolute;
-      right: 0;
-      bottom: 5px;
+      display: flex;
+      align-items: center;
       color: #888;
-      margin-left: 4px;
+      font-size: 14px;
```
oterral: This font is uselles since 14px is already apply.
Could it be possible to define the size of the i button in one place , see line 100. So .wkp-info-topic-bt and .wkp-info-layer-bt share the same width and height

edellucien:
@oterral I adapted the code again :)

PR #519, opened 2020-02-24
Fix spacing between layers and topics

edellucien:


src/components/TopicMenu/TopicMenu.scss
```diff
@@ -95,6 +95,7 @@
     margin: 2px 0 0 20px;
 
     .rs-layer-tree {
+      margin-top: -1px;
```
oterral: do we really need this one ?

edellucien: To ensure the pixel precise distance between layer items and topics

oterral: Why not putting margin-top: 1px to the wkp-layer-tree class instead dof 2px ?

oterral: maybe remove the wkp-layer-tree element , it seems useless.

edellucien: It works as well indeed

PR #522, opened 2020-02-25
Tracking

friedjoff:


src/components/TrafimageMaps/TrafimageMaps.js
```diff
@@ -91,8 +97,15 @@ const defaultProps = {
...
   language: 'de',
+  disableTracking: false,
 };
 
+const matomo = createInstance({
+  urlBase: process.env.REACT_APP_MATOMO_URL_BASE,
+  siteId: process.env.REACT_APP_MATOMO_SITE_ID,
+  trackerUrl: `${process.env.REACT_APP_MATOMO_URL_BASE}piwik.php`,
+});
+
```
oterral: Does it triggers errors if these variables are not defined ?

src/WebComponent.js
```diff
@@ -111,6 +116,7 @@ const attributes = {
 const defaultProps = {
   topics: undefined,
   history: undefined,
+  disableTracking: false,
```
oterral: should be true by default. I'm not sure if  we wants it activate or not for stelen

src/components/TopicLoader/TopicLoader.js
```diff
@@ -148,6 +150,8 @@ class TopicLoader extends Component {
       return;
     }
 
+    matomo.trackEvent({ category: activeTopic.name, action: 'load' });
```
oterral: if the creation of matomo object failed because there is no MATOMO_XX variables defined. what's happening here ?

src/WebComponent.js
mhaertwig: Agree. and I would call it `enableTracking`.

PR #526, opened 2020-02-27
Set font-size to 16px on input to avoid iphone to zoom on inputs

edellucien:


edellucien:
Font size changed to 16px to avoid safari to zoom on input, on click.

public/index.html
```diff
@@ -89,7 +96,7 @@
           });
 
         var map = document.getElementById('map');
-        var evts = document.getElementById('log');
+        const evts = document.getElementById('events');
```
oterral:
```suggestion
        var evts = document.getElementById('events');
```

src/components/Footer/Footer.scss
```diff
@@ -108,3 +108,13 @@
...
   }
 }
+
+.tm-w-xs {
+  .wkp-footer-left {
+    font-size: 11px;
+  }
+
+  .wkp-language-select {
+    padding: 1px 0;
```
oterral:
```suggestion
    padding-right: 0;
```

src/components/Search/Search.scss
```diff
@@ -10,7 +10,7 @@
   .wkp-search-button {
     border: none;
     color: white;
-    height: 41px;
+    height: 39px;
```
oterral: breaks on large screen the button doesn't touch the bottom border

src/components/Search/Search.scss
```diff
@@ -139,4 +139,12 @@
       margin: 0;
     }
   }
+
+  &.tm-w-xs,
+  &.tm-w-s {
+    .react-autosuggest__input {
+      font-size: 16px;
+      padding: 12px 15px 9px 15px;
```
oterral: try to find a padding that work for large and small screen if possible.

oterral: Not related to this PR but also I see the text oes under the cross and the red button.

src/components/MapControls/MapControls.scss
```diff
@@ -37,3 +37,19 @@
...
+  }
+}
+
+.tm-w-xs {
+  .rs-zooms-bar {
+    .ol-zoomslider {
+      height: 120px;
+    }
+  }
+}
```
oterral: the size of the zoom bar should be related to the  height not the width  `.tm-h-s`

src/components/Search/Search.scss
edellucien: The font-size change on mobile makes the input bigger, I have to adapt the padding, because we don't want to set the font-size to 16px on normal screens.

edellucien: I fixed the text going under the clear & search button additionally.

src/components/MapControls/MapControls.scss
edellucien: fixed

src/components/Search/Search.scss
edellucien: fixed

PR #527, opened 2020-02-28
Allow to show only one occurance of the same feature in different layers

edellucien:


mhaertwig:
The requirement is to disable popups for specific layers. This can be achieved by not defining a `popupComponent` for this layers.

PR #528, opened 2020-02-28
Display popup when we select a station in search results

oterral:
Also remove  sources mapbox layer, and fix update of stations sources behavior.


friedjoff:
Code looks good. In terms of UX it seems a little confusing to show the popup only once the layer is loaded. Some kind of spinner would be nice to tell the user something is going to happen. @oterral do you think thats possible?

oterral:
Problem is we have no choice than waiting for the map is loaded before displaying the popup. 
@friedjoff  where do you want to put a spinner ?

friedjoff:
The search result contains the coordinate, right? So we could display a popup with a spinner inside.

oterral:
Ah ok I thought a spinner on the map. I can try to see if I find something.

oterral:
I discuss with @mhaertwig . It's a good idea but not a high prority. So I merge .

PR #529, opened 2020-02-28
HidePopup for 'stuetzpunktBahnhoefe' if same feature already clicked

edellucien:


src/components/Popup/Popup.js
```diff
@@ -18,7 +18,7 @@ const Popup = () => {
 
     if (layer.get('popupComponent')) {
       if (typeof layer.hidePopup === 'function') {
-        return features.find(f => !layer.hidePopup(f));
+        return features.find(f => !layer.hidePopup(f, layer, featureInfo));
```
oterral: Is there already the layer in featureInfo object ?

edellucien: 'featureInformation' holds potentially several 'info', and 'layer' comes from the specifically tested 'info'.

PR #542, opened 2020-03-04
CasaStyle Updates

danji90:
- Separated Zone polygons from Zone Labels (layer group with 2 different layers)
- Adding Zones to both layers
- Applying different styles to the layers (unselected, selected and hover)
- Created setupTest file to define global test functions
- Updated zone tests
- Inserted new styleguidist example for CASA

### DIFF
--- a/package.json
+++ b/package.json
@@ -180,6 +180,9 @@
     ],
     "snapshotSerializers": [
       "jest-serializer-html"
+    ],
+    "setupFiles": [
+      "./setupTest.js"
     ]
   },
   "lint-staged": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "trafimage-maps",
-  "version": "1.0.7-beta.71",
+  "version": "1.0.7-beta.76",
   "main": "build/bundle.js",
   "proxy": "http://127.0.0.1:8000",
   "dependencies": {
@@ -19,7 +19,7 @@
     "i18next-xhr-backend": "^3.2.2",
     "jsts": "^2.1.0",
     "lodash": "^4.17.11",
-    "mapbox-gl": "^1.8.0",
+    "mapbox-gl": "^1.8.1",
     "node-sass": "^4.13.1",
     "ol": "6.1.1",
     "package.json": "^2.0.1",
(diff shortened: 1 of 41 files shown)


package.json
```diff
@@ -180,6 +180,9 @@
     ],
     "snapshotSerializers": [
       "jest-serializer-html"
+    ],
+    "setupFiles": [
+      "./setupTest.js"
```
oterral: if you call it setupTests.js and you move it to src/ folder .  CRA will find it automatically so you don't have to add it here.

src/layers/ZoneLayer/ZoneLayer.js
```diff
@@ -48,14 +49,28 @@ class ZoneLayer extends CasaLayer {
...
     super({
       name: 'Verbundzonen',
-      olLayer: new OLVectorLayer({
-        className: 'Verbundzonen', // needed for forEachLayerAtPixel
-        source: new VectorSource(),
-        style: (f, r) => this.zoneStyle(f, r),
+      olLayer: new LayerGroup({
+        layers: [
+          new OLVectorLayer({
+            className: 'Verbundzonen', // needed for forEachLayerAtPixel
```
oterral: What's so important with the className ? you never use it.

src/layers/ZoneLayer/ZoneLayer.js
```diff
@@ -218,7 +232,17 @@ class ZoneLayer extends CasaLayer {
...
+    this.featuresLayer.getSource().clear();
+    this.labelsLayer.getSource().clear();
+  }
+
+  /**
+   * Returns available layers in the layer group
+   * @private
+   * @inheritdoc
+   */
+  getLayers() {
```
oterral: Where do you use it ?

src/examples/Casa/README.md
```diff
@@ -59,13 +51,80 @@ zoneLayer.loadZones([
...
+      },
+    ],
+  },
+  {
+    partnerCode: 446,
+    zones: [
       {
         zoneCode: 170,
+        isSelected: false,
+        isClickable: true,
```
oterral: Is that wanted to specify isSelected and isClickable everywhere?

danji90: This was set in the file provided the SBB, not necessary though, since we have default values

src/layers/ZoneLayer/ZoneLayer.js
danji90: I kept in in the code to be able to access the layers from the layergroup from outside. I can remove it if it doesn't seem useful

src/layers/ZoneLayer/ZoneLayer.js
danji90: This was in the code before, I just copied it for the layers. It is actually important in the olLayer (removing will break the app), but in the layers I could probably remove it

package.json
danji90: ok

mhaertwig:
:+1: thanks

PR #554, opened 2020-03-09
Fix CASA route popups

danji90:
- Converting individual LineStrings in RoutesLayer to one MultiLineString on fetch
- Prevents multiple popup rendering
- Added functionality in Map component for showPopupOnHover attribute to filter the features inside the individual featureInfos as a function


src/components/Map/Map.js
```diff
@@ -128,10 +128,21 @@ class Map extends PureComponent {
...
-        infos = infos.filter(
-          ({ layer }) =>
-            layer.get('showPopupOnHover') && layer.get('popupComponent'),
-        );
+        infos = infos
+          .filter(
+            ({ layer }) =>
+              layer.get('showPopupOnHover') && layer.get('popupComponent'),
+          )
+          .map(i => {
```
oterral: call it `info`  we don' t understand what is `i`

src/components/Map/Map.js
```diff
@@ -128,10 +128,21 @@ class Map extends PureComponent {
...
+            /* Apply showPopupOnHover function if defined to further filter features */
+            const showPopupOnHover = i.layer.get('showPopupOnHover');
+            if (typeof showPopupOnHover === 'function') {
+              const newFeature = i;
+              newFeature.features = i.layer.get('showPopupOnHover')(i.features);
+              return newFeature;
```
oterral:
```
return {
   ...info,
   features: i.layer.get('showPopupOnHover')(i.features)
}
```

src/layers/RouteLayer/RouteLayer.js
```diff
@@ -26,7 +28,12 @@ class RouteLayer extends CasaLayer {
       }),
       ...options,
     });
-    this.set('showPopupOnHover', true);
+    this.set('showPopupOnHover', features => {
+      if (features.length) {
```
oterral: You don't need the `if` you can apply the filter on an empty array

src/layers/RouteLayer/RouteLayer.js
```diff
@@ -26,7 +28,12 @@ class RouteLayer extends CasaLayer {
       }),
       ...options,
     });
-    this.set('showPopupOnHover', true);
+    this.set('showPopupOnHover', features => {
```
oterral:
```suggestion
    this.set('showPopupOnHover', (features = []) => {
```
to make sure it's always an array

src/components/Map/Map.js
danji90: Done

PR #556, opened 2020-03-11
Lucien/gen levels

edellucien:


friedjoff:
grafik

edellucien:
On big zooms it seems inherent from the generalization levels we have in backend.
I will ask PO in the ticket.

edellucien:
PO decide to take care of the high  zooms in a future task.
Requirement here is only to fix generalization level 5 & 150.
Thanks for catching it anyway :)

PR #557, opened 2020-03-11
CASA hover style

danji90:
- Last adjustment to casa style
- added hover css on map controls in casa css


danji90:
@oterral this is a last small requirement from the casa style ticket. It only adds hover to the casa map controls. Mario already merged the other branch so I created this PR. Please merge if you agree

PR #558, opened 2020-03-11
Lucien/lazy loading topics

edellucien:
Fix info Button for intern topic


src/components/FeatureInformation/FeatureInformation.js
```diff
@@ -17,6 +17,7 @@ const propTypes = {
 const FeatureInformation = ({ featureInfo }) => {
   const { t } = useTranslation();
   const dispatch = useDispatch();
+  const language = useSelector(state => state.app.language);
```
mhaertwig: @edellucien i would just use the `t()` function in the popup component.

edellucien: The language is relevant for some topics because the field that needs to be displayed in the popup are langauge specific:
'text_de', 'text_fr', 'text_it', which then requires to know the current language.

PR #560, opened 2020-03-12
Enhance accessiblity with aria-labels

edellucien:


friedjoff:
@edellucien looks fine! I'll let you merge, because I don't know if this is still WIP.

PR #569, opened 2020-03-16
IOs CSS fixes

danji90:


src/components/TrafimageMaps/TrafimageMaps.scss
```diff
@@ -141,7 +141,7 @@
     &.header {
       .wkp-menu-wrapper {
         top: 0;
-        width: 345px;
+        width: 100vw;
```
oterral: This one breaks the display on window's width: 960px . The menu is way too big.

src/components/Search/Search.scss
```diff
@@ -2,38 +2,41 @@
...
   right: 320px;
   left: 400px;
+  z-index: 2; // To be above header div, in order to be clickable.
 
   .wkp-search-button {
     border: none;
     color: white;
-    height: 40px;
-    width: 46px;
+    height: calc(#{$header-height} - 64px);
```
oterral: if 64px is the height of the input put it also in variable.

oterral:
@danji90  if you have time . These css are really easy to break. Can you try to use the `renderInputComponent` attribute of Autosuggest ? 
```
const renderInputComponent = inputProps => (
  <div className="input-container>
    <input {...inputProps} />
    <button>the clear buttonr</button>
    <button>the searcgh button</button>
  </div>
);
```

so with a display flex you could be able manage the height of clear and search button better.

oterral:
https://github.com/moroshko/react-autosuggest

src/components/TrafimageMaps/TrafimageMaps.scss
danji90: @oterral I just tested on all kinds of disply sizes and it seems to work for all cases. 100vh takes 100% of the view port width no? This was what fixed the css on the iPhone 5 display, since the menu wrapper was overflowing

danji90:
Recreated branch and created new pull request

PR #570, opened 2020-03-16
Downgrade mapbox-gl to v1.6.0

edellucien:


yarn.lock
```diff
@@ -11019,7 +11019,7 @@ map-visit@^1.0.0:
   dependencies:
     object-visit "^1.0.0"
 
-mapbox-gl@^1.8.1:
+mapbox-gl@^1.6.0:
   version "1.8.1"
```
oterral: @edellucien  that's not what I call a downgrade . ;), you re still using 1.8.1.

edellucien: Indeed sorry did that too fast

PR #573, opened 2020-03-17
Fix handicap overlays in order to fit to the basemap

gevos:
Please, have a look at the commit message of this [commit](https://github.com/geops/trafimage-maps/pull/573/commits/e40bb061e0e3ba5dded95633e789606a28fccf38)

# How to

Test branch with dev data: https://github.com/geops/trafimage-maps/pull/624
Should also work with prod data (preview app in this PR)

Then switch to handicap topic

# Others

- [x] It's not a hack or at least an unauthorized hack :).
- [x] Everything in ticket description has been fixed.
- [x] The author of the MR has made its own review before assigning the reviewer.
- [x] IE11 tested.
- [x] The title means something for a human being.
- [ ] Labels applied. if it's a release? a hotfix?
- [ ] Tests added.

PR #574, opened 2020-03-17
Line search: Fix highlight style

gevos:


gevos:
Commit comes from https://github.com/geops/trafimage-maps/pull/573

edellucien:
:+1:

gevos:
@oterral Could you have a look at it and merge it? I think, it's a small bugfix where we do not need to ask the PO.

PR #577, opened 2020-03-17
iPhone & Search fix

danji90:
- Redefined base layer dropdown width (preventing container overflow)
- Using renderInputComponent prop for AutoSuggest, using flexBox to improve CSS
- Removed setSearchOpen dispatch on search clear button
- Improved Search component centering
- Removed Developer Portal link from footer on small screens

### DIFF
--- a/src/components/Footer/Footer.scss
+++ b/src/components/Footer/Footer.scss
@@ -112,6 +112,10 @@
 .tm-w-xs {
   .wkp-footer-left {
     font-size: 11px;
+
+    .wkp-dev-portal-link {
+      display: none;
+    }
   }
 
   .wkp-language-select {
--- a/src/components/Search/Search.js
+++ b/src/components/Search/Search.js
@@ -1,3 +1,4 @@
+/* eslint-disable react/jsx-props-no-spreading */
 import React, { useEffect, useMemo, useRef, useState } from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 import Autosuggest from 'react-autosuggest';
@@ -21,6 +22,40 @@ function Search() {
   const searchService = useSelector(state => state.app.searchService);
   const { t } = useTranslation();
   const dispatch = useDispatch();
+  const renderInputComponent = inputProps => {
+    return (
+      <div className="wkp-search-input">
+        <input {...inputProps} />
+        {value && (
+          <button
+            type="button"
+            tabIndex={0}
+            aria-label={t('Suchtext löschen')}
+            className="wkp-search-button wkp-search-button-clear"
+            onClick={() => {
+              setValue('');
+              searchService.clearHighlight();
+            }}
+          >
+            <FaTimes />
+          </button>
+        )}
+        <button
+          type="button"
+          tabIndex={0}
+          aria-label="Suche"
+          className="wkp-search-button wkp-search-button-submit"
+          onClick={() => {
+            if (!value) {
+              dispatch(setSearchOpen(false));
+            }
+          }}
+        >
+          <FaSearch focusable={false} />
+        </button>
+      </div>
+    );
+  };
 
   const componentIsMounted = useRef(true);
   useEffect(() => {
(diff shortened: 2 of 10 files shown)


src/components/Search/Search.js
```diff
@@ -21,6 +22,40 @@ function Search() {
   const searchService = useSelector(state => state.app.searchService);
   const { t } = useTranslation();
   const dispatch = useDispatch();
+  const renderInputComponent = inputProps => {
+    return (
+      <div className="wkp-search-input">
+        <input {...inputProps} />
```
danji90: Eslint complains about not spreading props, but this is the way React-AutoSuggest defines. Maybe there is a better way though

src/components/Search/Search.scss
```diff
@@ -2,40 +2,47 @@
...
+      align-items: center;
+      border: 2px solid #b7b7b7;
+    }
+  }
+
   .wkp-search-button {
     border: none;
     color: white;
-    height: 40px;
+    height: 42px;
```
oterral: now you can use height:100%  isn't it ?

src/components/Search/Search.scss
```diff
@@ -2,40 +2,47 @@
...
 
     &.wkp-search-button-clear {
-      right: 44px;
       background: none;
       color: #666;
     }
 
     &.wkp-search-button-submit {
       background-color: $brand-primary;
-      right: -2px;
```
oterral: :clap:

src/components/Search/Search.scss
```diff
@@ -124,6 +132,17 @@
...
+    .wkp-search-toggle-container--open {
+      .wkp-search-input {
+        display: flex;
+        justify-content: center;
+        align-items: center;
+        border: 2px solid #b7b7b7;
+        background-color: white;
+        width: 100%;
+      }
+    }
```
oterral: What is different with lines from line 10 ?

src/components/Search/Search.scss
```diff
@@ -140,21 +159,18 @@
...
 
   &.tm-w-xs,
   &.tm-w-s {
     .wkp-search-button {
-      height: 39px;
+      width: 40px;
     }
 
     .react-autosuggest__input {
+      width: calc(100% - 90px);
```
oterral: If you use display flex the width should be automatic calculated.

src/components/Search/Search.js
oterral: no it's ok for this one. you can deactivate it

src/components/Footer/Footer.scss
```diff
@@ -112,6 +112,10 @@
 .tm-w-xs {
   .wkp-footer-left {
     font-size: 11px;
+
+    .wkp-dev-portal-link {
+      display: none;
+    }
```
oterral: The owner of the developer portal will be disappointed if we remove this link? is that ok with @mhaertwig  ?

src/components/Search/Search.scss
danji90: Yes, this can be simplified. will do

src/components/Search/Search.scss
danji90: The problem here is that the clear button is inserted dynamically and with the larger text font the container content will overflow once there is an input value

src/components/Footer/Footer.scss
danji90: @mhaertwig said we can remove it on small resolutions, since the footer becomes really cramped and the content won't fit

src/components/Search/Search.scss
danji90: I fixwd this in the meantime by making the button flex containers with height:100%, so the svgs align without padding/margin

src/components/Search/Search.scss
```diff
@@ -2,47 +2,58 @@
...
 .wkp-search {
   position: absolute;
-  top: 25px;
+  top: 26px;
   z-index: 2; // To be above header div, in order to be clickable.
   right: 320px;
   left: 400px;
 
+  .wkp-search-toggle-container--open,
+  .wkp-search-toggle-container {
```
oterral: Do you need this 2 lines ?

danji90: You're right, they're useless. Thanx for pointing that out. Fixed and pushed

PR #580, opened 2020-03-18
Only add http to link href when missing initially

edellucien:
Add http to handicap href link only when not initially present.


src/popups/HandicapPopup/HandicapPopupElement.js
```diff
@@ -55,10 +55,11 @@ const telTo = (telNumber, idx) => (
   </a>
 );
 
-const urlHref = (href, idx) => {
+const urlHref = (link, idx) => {
+  const href = /(http(s?)):\/\//i.test(link) ? link : `http://${link}`;
```
oterral:
```suggestion
  const href = /^http(s?):\/\//i.test(link) ? link : `http://${link}`;
```

PR #581, opened 2020-03-18
rolling back custom-component lib to 1.3.2

danji90:
* Downgraded @webcomponents/custom-elements from 1.4.0 to 1.3.2


danji90:
Tested successfully on IE 11 and Edge

PR #616, opened 2020-03-31
Lucien/topic collapser

edellucien:
# How to

Hide collapse arrow if no visible layers are to be toggled.

# Others

- [x] IE11 tested.
- [ ] Tests added.

### DIFF
--- a/src/components/TopicMenu/TopicMenu.js
+++ b/src/components/TopicMenu/TopicMenu.js
@@ -189,7 +189,7 @@ class TopicMenu extends PureComponent {
     const collapsed = isCollapsed || activeTopic.key !== topic.key;
 
     const menuVisibleLayers = (topic.layers || []).filter(l => {
-       return !l.get('hideInLegend') && !l.isMapboxStyleLayer;
+      return !l.get('hideInLegend') && !l.isMapboxStyleLayer;
     });
     return (
       <div className="wkp-topic-menu">
--- a/src/components/TopicMenu/TopicMenu.js
+++ b/src/components/TopicMenu/TopicMenu.js
@@ -9,6 +9,7 @@ import Select from '@geops/react-ui/components/Select';
 import LayerService from 'react-spatial/LayerService';
 import Button from '@geops/react-ui/components/Button';
 import Layer from 'react-spatial/layers/Layer';
+import MapboxStyleLayer from '../../layers/MapboxStyleLayer';
 import Collapsible from '../Collapsible';
 import { setActiveTopic, setSelectedForInfos } from '../../model/app/actions';
 
@@ -188,6 +189,11 @@ class TopicMenu extends PureComponent {
 
     const collapsed = isCollapsed || activeTopic.key !== topic.key;
 
+    const menuVisibleLayers = !topic.layers
+      ? []
+      : topic.layers.filter(l => {
+          return !l.get('hideInLegend') && !(l instanceof MapboxStyleLayer);
+        });
     return (
       <div className="wkp-topic-menu">
         <div className="wkp-topic-menu-item-wrapper">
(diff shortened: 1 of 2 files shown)


src/layers/MapboxStyleLayer/MapboxStyleLayer.js
```diff
@@ -40,6 +40,7 @@ const applyLayoutVisibility = (mbMap, visible, filterFunc) => {
 class MapboxStyleLayer extends Layer {
   constructor(options = {}) {
     super(options);
+    this.isMapboxStyleLayer = true;
```
edellucien: use duckTyping because instanceof will not work with lazy loading of intern topics

src/components/TopicMenu/TopicMenu.js
```diff
@@ -188,6 +188,11 @@ class TopicMenu extends PureComponent {
 
     const collapsed = isCollapsed || activeTopic.key !== topic.key;
 
+    const menuVisibleLayers = !topic.layers
+      ? []
+      : topic.layers.filter(l => {
+          return !l.get('hideInLegend') && !l.isMapboxStyleLayer;
+        });
```
oterral, line 195:
```suggestion
    const menuVisibleLayers = (topic.layers || []).filter(l => {
       return !l.get('hideInLegend') && !l.isMapboxStyleLayer;
    });
```

src/layers/MapboxStyleLayer/MapboxStyleLayer.js
oterral: Why don't  you specify hideInLegend prop for the layer you want to hide in legend ?

edellucien: I thought we would never want mapboxStyle Layer to appear in the layerTree.
I can add hideInLegend to all of them if it makes more sense.

oterral: There is a lot of MapboxStyle Layer in the layer tree, see netzkarte topic

oterral: and eisenbahninfrastruktur

edellucien: That's why I wanted to hide them all by default :smiley:

edellucien: This is fixed, for infofpw, we will fix it on gitlab

src/components/TopicMenu/TopicMenu.js
```diff
@@ -207,12 +210,14 @@ class TopicMenu extends PureComponent {
               </div>
               {t(topic.name)}
             </div>
-            <div
-              className={`wkp-layer-toggler ${collapsed ? 'collapsed' : ''}`}
-              style={{
-                display: topic.key === activeTopic.key ? 'block' : 'none',
-              }}
-            />
+            {menuVisibleLayers.length ? (
```
oterral, line 213: I discover this works:
```suggestion
            {!!menuVisibleLayers.length && (
```

src/components/TopicMenu/TopicMenu.js
```diff
@@ -188,6 +188,9 @@ class TopicMenu extends PureComponent {
 
     const collapsed = isCollapsed || activeTopic.key !== topic.key;
 
+    const menuVisibleLayers = (topic.layers || []).filter(l => {
+      return !l.get('hideInLegend') && !l.isMapboxStyleLayer;
```
oterral, line 192:
```suggestion
      return !l.get('hideInLegend');
```

oterral: You don't need to filter just do a find , It will stop the loop when he find the first hideInLegend=false.

edellucien: Good idea use find instead of filter, I implemented

src/components/TopicMenu/TopicMenu.js
```diff
@@ -207,12 +210,14 @@ class TopicMenu extends PureComponent {
...
-              }}
-            />
+            {isMenuVisibleLayers ? (
+              <div
+                className={`wkp-layer-toggler ${collapsed ? 'collapsed' : ''}`}
+                style={{
+                  display: topic.key === activeTopic.key ? 'block' : 'none',
+                }}
+              />
+            ) : null}
```
oterral, line 220: As I said in a previous comment when it's a boolean you can do : 
```suggestion
            {isMenuVisibleLayers && (
              <div
                className={`wkp-layer-toggler ${collapsed ? 'collapsed' : ''}`}
                style={{
                  display: topic.key === activeTopic.key ? 'block' : 'none',
                }}
              />
            )}
```

oterral:
just a small suggestion.

PR #627, opened 2020-04-01
WIP Add Manual link in ShareMenu

edellucien:
# How to

Add quickstart manual link for intern topics

# Others

- [x] IE11 tested.
- [x] Tests added.

### DIFF
--- a/src/components/Share/Share.js
+++ b/src/components/Share/Share.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
 import { TiImage, TiSocialFacebook, TiSocialTwitter } from 'react-icons/ti';
-import { FaEnvelope, FaPencilAlt } from 'react-icons/fa';
+import { FaEnvelope, FaPencilAlt, FaQuestion } from 'react-icons/fa';
 import CanvasSaveButton from 'react-spatial/components/CanvasSaveButton';
 import BlankLink from '@geops/react-ui/components/BlankLink';
 import Button from '@geops/react-ui/components/Button';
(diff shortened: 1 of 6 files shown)


edellucien:
@oterral Can you review and test the current state?
The quickstart Link is supposed to be displayed only for intern topic. But for you to test i display it for public topic first.
Then when you accept the MR I will switch the display for internal and edit the tests.

src/components/Share/Share.js
```diff
@@ -48,26 +54,36 @@ const renderConf = (conf, t) => (
...
+    if (socialShareConfig[i].url) {
+      const test = { ...duplicateConfig[i] }.url;
 
-  for (let i = 0; i < config.length; i += 1) {
-    if (config[i].url) {
-      config[i].url = config[i].url.replace('{url}', window.location.href);
+      duplicateConfig[i].url = test
+        .replace('{url}', window.location.href)
+        .replace('{language}', language)
+        .replace('{appBase}', appBaseUrl);
```
oterral, line 72: Why not {appBaseUrl} ?

src/components/Share/Share.js
```diff
@@ -48,26 +54,36 @@ const renderConf = (conf, t) => (
 );
 
 const Share = ({ appBaseUrl }) => {
+  const activeTopic = useSelector(state => state.app.activeTopic);
+  const language = useSelector(state => state.app.language);
   const map = useSelector(state => state.app.map);
   const { t } = useTranslation();
-  const config = [...socialShareConfig];
+  const duplicateConfig = JSON.parse(JSON.stringify(socialShareConfig));
```
oterral, line 61: Is that possible to have duplicate config ? why kind of person does that?

edellucien: without duplicating, we have initial language de.
We replace it in the url and have: 'url/de/...'
and if we change the language to 'it', replacing {lang} will not work anymore because it is already set ad 'de'.
and this won't update, because without duplicating we refer to the same object.

edellucien: So we always use the initial 'socialShareConfig' to replace, and our duplicate is not just a ref to the 'socialShareConfig'

src/components/Share/Share.js
edellucien: I changed to 'appBaseUrl'

edellucien:
Closed in favour of https://github.com/geops/trafimage-maps/pull/633
For base of PR to release.1.0.8

PR #635, opened 2020-04-02
Increase max height info dialog on mobile

edellucien:
# How to

Increase max height info dialog on mobile
# Others

- [x] IE11 tested.
- [ ] Tests added.


src/components/Dialog/Dialog.scss
```diff
@@ -100,6 +100,7 @@
 
         .tm-dialog-body {
           height: calc(100% - 50px);
+          padding-bottom: 0;
```
oterral, line 103: You shouldn't have to do this, there is a missing  overflow: hidden somewhere

PR #638, opened 2020-04-03
[WIP] Gevos/zweitausbildung patches

gevos:
# How to

Additional changes for https://github.com/geops/trafimage-maps/pull/453 which have not been reviewed.

[Test application](https://deploy-preview-639--dev-trafimage-maps.netlify.com) (netlify from https://github.com/geops/trafimage-maps/pull/639) as the topic is hidden in the LayerTree, now

# Others

- [x] Everything in ticket description has been fixed. - Otherwise, I will open another PR.
- [x] IE11 tested.
- [x] Tests added. - No, probably not needed

### DIFF
--- a/src/components/FeatureInformation/FeatureInformation.scss
+++ b/src/components/FeatureInformation/FeatureInformation.scss
@@ -25,13 +25,15 @@
   }
 
   .wkp-feature-information-body {
-    padding: 10px;
+    > div {
+      padding: 10px;
+    }
 
     .wkp-pagination-wrapper {
       display: flex;
       align-items: center;
       justify-content: center;
-      padding-top: 5px;
+      padding-top: 0;
 
       .wkp-pagination-button-wrapper {
         padding: 0 10px;
--- a/src/components/IconList/IconList.js
+++ b/src/components/IconList/IconList.js
@@ -124,7 +124,17 @@ class IconList extends PureComponent {
       }
 
       return ReactDOM.createPortal(
-        <div className="wkp-icon-list-modal">
+        <div
+          role="button"
+          tabIndex={0}
+          className="wkp-icon-list-modal"
+          onClick={() => {
+            this.setState({
+              iconListVis: false,
+            });
+          }}
+          onKeyPress={() => {}}
+        >
           <div className="wkp-icon-list-modal-pos" style={listStyles}>
             {list}
           </div>
--- a/src/i18n.js
+++ b/src/i18n.js
@@ -39,6 +39,7 @@ i18n
         escapeValue: false,
       },
       keySeparator: ' µ', // Deactivate the keySeparator option
+      nsSeparator: false, // Deactivate the nsSeparator option
       saveMissing: false,
       missingKeyHandler(lng, ns, key, fallbackValue) {
         // eslint-disable-next-line no-console
--- a/src/popups/ZweitausbildungPoisPopup/ZweitausbildungPoisPopup.scss
+++ b/src/popups/ZweitausbildungPoisPopup/ZweitausbildungPoisPopup.scss
@@ -1,7 +1,6 @@
 .wkp-zweitausbildung-pois-popup {
   overflow-y: auto;
   max-height: 300px;
-  margin-right: -10px;
 
   .wkp-zweitausbildung-pois-popup-row {
     min-height: 50px;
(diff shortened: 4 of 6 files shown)


src/i18n.js
```diff
@@ -39,6 +39,7 @@ i18n
         escapeValue: false,
       },
       keySeparator: ' µ', // Deactivate the keySeparator option
+      nsSeparator: false, // Deactivate the nsSeparator option
```
gevos, line 42: Core change

oterral: Why ?

src/popups/ZweitausbildungPoisPopup/ZweitausbildungPoisPopup.scss
```diff
@@ -1,6 +1,8 @@
 .wkp-zweitausbildung-pois-popup {
   overflow-y: auto;
   max-height: 300px;
+  padding-right: 10px;
+  margin-right: -10px;
```
oterral, line 5: That feels like a hack :D

gevos: Maybe, it can be regarded as an hack. I can't access the surrounding div element in order to remove the padding for the scrollbar as I would also break all other popups.

oterral: as I see the style is already broken if we have to do that. Let me try something

src/i18n.js
gevos: In the special case of the topic zweitausbildung, we have a dropdown item 'Alle' which should be translated. This is currently translated in the general IconList component, see [line|https://github.com/geops/trafimage-maps/blob/gevos/zweitausbildung-patches/src/components/IconList/IconList.js#L72]. Other dropdown items have names that contain colons, such as "Bernina Express: Chur - ...". These items do not have to translations but are translated anyways.

In the default behavior of i18n a text containing a colon will be split. The first part is regarded as namespace, the second part is returned. As far as I know, we don't use namespaces in this application. So, I thought it would be safe to exclude them. Then we don't have to think about colons in texts which I think is a trap.

src/popups/ZweitausbildungPoisPopup/ZweitausbildungPoisPopup.scss
gevos: Alternatively, we could set a max-height for the general popup with a scrollbar.

oterral: @gevos  try with this : https://github.com/geops/trafimage-maps/pull/640

gevos: I have cherrypicked your commit.

src/components/FeatureInformation/FeatureInformation.scss
```diff
@@ -25,13 +25,15 @@
   }
 
   .wkp-feature-information-body {
-    padding: 10px;
+    > div {
+      padding: 10px;
+    }
```
gevos, line 30: Core change

src/components/FeatureInformation/FeatureInformation.scss
```diff
@@ -25,13 +25,15 @@
...
 
     .wkp-pagination-wrapper {
       display: flex;
       align-items: center;
       justify-content: center;
-      padding-top: 5px;
+      padding-top: 0;
```
gevos, line 36: Core change

src/components/IconList/IconList.js
```diff
@@ -124,7 +124,17 @@ class IconList extends PureComponent {
...
+          role="button"
+          tabIndex={0}
+          className="wkp-icon-list-modal"
+          onClick={() => {
+            this.setState({
+              iconListVis: false,
+            });
+          }}
+          onKeyPress={() => {}}
+        >
```
gevos, line 137: Core change

> but we also can't go out of this modal, unless we click on the select box. It's annoying. Can't you make the modal bg transparent and on a user interaction on it, you close the select box ?

from https://github.com/geops/trafimage-maps/pull/453#discussion_r402965394

I have added an onClick listener which closes the dropdown.

However, the modal dropdown cannot be used/closed by keys. As it is appended to the .tm-trafimage-maps element we would have to manually change the focus. With an onFocus callback function for the react-ui Button component we could do this. @oterral, what do you think?

gevos: > With an onFocus callback function for the react-ui Button component we could do this.

@oterral, The onFocus callback function would not solve this problem. I have to react on the onClick function

gevos: We won't support barrier free behavior, for now.

gevos:
As @oterral has approved this PR, I will merge it.

PR #644, opened 2020-04-06
Fix Cypress tests

danji90:


oterral:
@danji90  it's good cypress pass but all the rest failed :D

oterral:
Thxx!!!!

PR #646, opened 2020-04-07
Ability to pass custom resolution/generalisation

edellucien:
# How to

Allow to pass data specific resolution array to LaterHelper function.
Ex: Lines geom in the search and mobz data have different resolutions available.

# Others

- [x] IE11 tested.
- [x] Tests added.


edellucien:
Reset resolution/generalisation values to default.
Add ability to pass custom resolution/generalisation values.

src/layers/layerHelper.js
```diff
@@ -1,18 +1,23 @@
-const dataResolutions = [900, 850, 500, 250, 115, 100, 75, 50, 20, 10, 5];
+const dataResolutions = [750, 500, 250, 100, 75, 50, 20, 10, 5];
 
-function getDataResolution(resolution) {
-  return dataResolutions.reduce((prev, curr) =>
+function getDataResolution(resolution, resolutions) {
```
oterral, line 3:
```suggestion
function getDataResolution(resolution, resolutions = dataResolutions) {
```

src/layers/layerHelper.js
```diff
@@ -1,18 +1,23 @@
...
+  return (resolutions || dataResolutions).reduce((prev, curr) =>
```
oterral, line 4:
```suggestion
  return (resolutions.reduce((prev, curr) =>
```

edellucien:
Thanks for the fast feedback!
I will apply the suggestions! Make sense to use parameter default value indeed!

PR #658, opened 2020-04-22
Zweitausbildung: Restrict max zoom to 13

gevos:
# How to

- Switch to topic ch.sbb.zweitausbildung (Development button -> change topic)
- Zoom in until the zooming stops
- The z value in the permalink should be 13

# Others

- [x] IE11 tested.


gevos:
As it is only a small change, I will merge it on my own.

gevos:
Zweitausbildung: Restrict zoom to 13 (ca. resolution 20)...
    
...which corresponds to the restriction in the old WKP (zoom 5)

PR #668, opened 2020-04-30
Overlay correction for v3 style

gevos:
# How to

1. Use test PR https://github.com/geops/trafimage-maps/pull/669 to test with up-to-date dev data
2. Go to topic behig
3. Check for different zoom levels (especially 7 and below) if the overlays are displayed when a station label of the basemap is displayed. 
Check also the overlays fit for decimal point zoom levels (click on Ganze Schweiz button and zoom)

(There might be exceptions where a station label is missing due to collisions in the basemap style.)

# Others

- [x] IE11 tested.
- [ ] Tests added. - Not needed


edellucien:
LGTM
Very good work, thanks a lot for the testcase explanation, very handy to test!
:tada:

PR #673, opened 2020-05-06
Set apiKey for layers in TopicLoader...

gevos:
...This fixes the bug where the apiKey was not set for the punctuality layers (TrajservLayer)
when the existing topics where used without calling getTopicConfig() again
(see WebComponent.js line 174).

For the complete bugfix, please use the latest react-transit version
with getter/setter for apiKey.

# How to

1. Checkout this branch and create a build
2. Checkout trafimage-maps from GitLab and add this build 
3. Run the application
4. Open the developer tools
5. Make the punctuality layer visible
6. Check if the trajectory_collection request in the Network tab has a valid apiKey, e.g.
`https://api.geops.io/tracker/v1/trajectory_collection?...&key=5cc...a9aa400`,
`https://api.geops.io/tracker/v1/trajectory_collection?...&key=undefined` is not allowed (was the bug).

# Others

- [ ] IE11 tested. - Not possible
- [ ] The title means something for a human being.
- [ ] Tests added.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "react-shadow": "^17.6.0",
     "react-spatial": "0.2.14",
     "react-styleguidist": "^11.0.4",
-    "react-transit": "^0.3.2-beta.1",
+    "react-transit": "^0.3.2",
     "react-web-component": "^2.0.1",
     "redux": "^4.0.5",
     "redux-debounced": "^0.5.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -36,7 +36,7 @@
     "react-shadow": "^17.6.0",
     "react-spatial": "0.2.14",
     "react-styleguidist": "^11.0.4",
-    "react-transit": "^0.3.1",
+    "react-transit": "^0.3.2-beta.1",
```
gevos, line 39: Before merging, use a patch version of react-transit as soon as https://github.com/geops/react-transit/pull/280 is merged.

PR #676, opened 2020-05-07
Handicap popup: Restrict links to max width and use text-overflow ellipsis

gevos:
# How to

1. Open topic handicap
2. Open popup for Bern
3. The long link should overlap the popup border. Instead text-overflow ellipsis is used (...)
4. Other links in this popup should be displayed as before.

# Others

- [x] IE11 tested.
- [ ] Tests added. - I think, not needed.

### DIFF
--- a/src/popups/HandicapPopup/HandicapPopup.scss
+++ b/src/popups/HandicapPopup/HandicapPopup.scss
@@ -6,6 +6,18 @@
     text-decoration: underline;
   }
 
+  a.wkp-link {
+    text-decoration: none !important;
+  }
+
+  .wkp-link span {
+    text-decoration: underline;
+    text-overflow: ellipsis;
+    max-width: 380px;
+    overflow: hidden;
+    display: inline-block;
+  }
+
   .wkp-handicap-popup-body {
     .wkp-handicap-popup-title {
       font-weight: bold;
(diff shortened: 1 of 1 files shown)


src/popups/HandicapPopup/HandicapPopup.scss
```diff
@@ -6,6 +6,18 @@
...
+  a.wkp-link {
+    text-decoration: none !important;
+  }
+
+  .wkp-link span {
+    text-decoration: underline;
+    text-overflow: ellipsis;
+    max-width: 380px;
+    overflow: hidden;
+    display: inline-block;
```
gevos, line 18: With `display: inline-block;` the barrierfree focus box isn't perfect as the height at the arrow is smaller than at the text.
`display: table-cell;` could solve it. But it is [not supported by Opera Mobile](https://caniuse.com/#search=table-cell). If we don't want to ignore Opera Mobile, we could additionally add `display: table-cell;` after `display: inline-block;`. Then, I think, all browser would use `display: table-cell;` except for Opera Mobile. @edellucien what do you think?

edellucien: Set the arrow svg height, the same height as the text to have a correct highlight.
eg. 16px

gevos: Solved with the suggestion of https://github.com/geops/trafimage-maps/pull/676#discussion_r422097480

PR #683, opened 2020-05-14
CASA routes start/end style

danji90:
# How to

Routes in CASA layer should have a start and end style as in https://sbb.invisionapp.com/share/B6W8O98CEYK#/screens when selected (TRAFCASA-56).
The styles were added in CasaLayer.js in getOLStylesFromObject()
This can be tested in the netlify review app in CasaLayer  

# Others

- [x] IE11 tested.
- [x] Tests added.


src/layers/CasaLayer/CasaLayer.js
```diff
@@ -57,6 +59,13 @@ import ArrowImg from '../../img/arrow.png';
  * @param {styleFunction} [options.styleFunction] Style function.
  */
 class CasaLayer extends VectorLayer {
+  static isRoute = (geometry) => {
```
mhaertwig, line 62: Please do the route styling in the route layer and not inside the generic `CasaLayer`. The route styling doesn't have to be generic. Moreover, for readability I would not use an extra function for this simple `if` condition.

src/layers/CasaLayer/CasaLayer.js
```diff
@@ -57,6 +59,13 @@ import ArrowImg from '../../img/arrow.png';
...
+    /* Returns true if the geometry is a route */
+    const route =
```
mhaertwig, line 64: directly return.

src/layers/CasaLayer/CasaLayer.js
```diff
@@ -157,6 +166,42 @@ class CasaLayer extends VectorLayer {
     });
 
     const olStyles = {};
+    if (feature && CasaLayer.isRoute(feature.getGeometry())) {
+      /* Set route start and end style on selected routes */
+      if (isSelected) {
+        const circleStyle = (coords) =>
```
mhaertwig, line 172: Since it's a getter function, call it `getCircleStyle()` and move it out of the styling function. Otherwise the function is newly created on every call of the styling function. That's bad for performance.

src/layers/CasaLayer/CasaLayer.js
danji90: I did the styling in  CASA Layer, since all the other styles are defined here as well, including route styles (like the arrows for instance). I will move all the styling into route/zone layers respectively.

danji90: I created the isRoute function to actually improve readability, since the if condition would be quite long and is used multiple times:
geometry instanceof LineString || geometry instanceof MultiLineString

src/layers/CasaLayer/CasaLayer.js
danji90: ok

src/layers/CasaLayer/CasaLayer.js
danji90: Good point

src/layers/CasaLayer/CasaLayer.js
mhaertwig: The generic styling functionality is done in the casa layer. Route styling should be done in the route layer.

PR #705, opened 2020-06-08
Test header elements with cypress

gevos:
# How to

Cypress test for header should successfully run

# Others

- [x] Tests added.

### DIFF
--- a/cypress/integration/Header.spec.js
+++ b/cypress/integration/Header.spec.js
@@ -1,0 +1,110 @@
+const viewports = [
+  // Desktop
+  {
+    size: [1440, 900],
+  },
+  // Large mobile phone
+  {
+    preset: 'samsung-s10',
+    orientation: 'portrait',
+  },
+  {
+    preset: 'samsung-s10',
+    orientation: 'landscape',
+  },
+  // Small mobile phone
+  {
+    preset: 'iphone-3',
+    orientation: 'portrait',
+  },
+  {
+    preset: 'iphone-3',
+    orientation: 'landscape',
+  },
+  // Tablet
+  {
+    preset: 'ipad-2',
+    orientation: 'portrait',
+  },
+  {
+    preset: 'ipad-2',
+    orientation: 'landscape',
+  },
+];
+
+const setViewPort = (viewport) => {
+  if (viewport.size) {
+    cy.viewport(viewport.size[0], viewport.size[1]);
+  } else {
+    cy.viewport(viewport.preset, viewport.orientation);
+  }
+};
+
+const getViewPortName = (viewport) => {
+  if (viewport.size) {
+    return viewport.size;
+  }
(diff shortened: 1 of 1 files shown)


cypress/integration/Header.spec.js
```diff
@@ -0,0 +1,110 @@
...
+        // Do not test click because of CORS problem
+        // when redirecting to another page
+      });
+
+      it(`should display language select`, () => {
+        setViewPort(viewport);
+        cy.get('.wkp-single-value-wrapper').should('be.visible');
+
+        // Click the react-select element
+        cy.get('.wkp-header .css-m869aq-ValueContainer').click();
```
edellucien, line 80: Isn't this class name gonna change on build? Maybe with an id we could select the correct item, else it is good enough.

cypress/integration/Header.spec.js
```diff
@@ -0,0 +1,110 @@
...
+        cy.get('.wkp-topics-menu-body').should('not.be.visible');
+        cy.get('.wkp-menu-header').click();
+
+        // Menu should open
+        cy.get('.wkp-topics-menu-body').should('be.visible');
+      });
+
+      it(`should display WKP login`, () => {
+        setViewPort(viewport);
+        cy.get('.wkp-login').should('be.visible');
```
edellucien, line 70: We should either find a way of testing the redirect or simply not write the test at all I think.

gevos: I have checked this. These [code samples](https://github.com/geops/trafimage-maps/blob/master/src/components/Login/Login.js#L19) show solutions how to test redirects. As the Login component does not have a href attribute and opens the new page by using window.location.href (see [line](https://github.com/geops/trafimage-maps/blob/master/src/components/Login/Login.js#L19)), we could only use [third option](https://github.com/cypress-io/cypress-example-recipes/blob/master/examples/testing-dom__tab-handling-links/cypress/integration/tab_handling_anchor_links_spec.js#L183) which sets chromeWebSecurity to false. However, when I do this, I get redirected to chrome-error://chromewebdata/. I could try to solve this. But, I think, the main idea of cypress is to test a single domain. 

I would prefer to only check if this Login button is available. It's better than checking nothing.

cypress/integration/Header.spec.js
gevos: @edellucien has found a workaround for the IABP project where a ValueContainer and Menu is defined for the react-select component, additionally to the SingleContainer. So, we have  beautiful css names. Unfortunately, this does not work in trafimage-maps. And we don't know why. We get a detached-from-DOM error. And even, if we do not change the classes in the test, it does not work anymore.

I have checked if the strange react-select css classes change after a built or when the react-select version changes - no. So, I prefer to leave this test the way it is. @edellucien agrees.

edellucien:
Thanks a lot for the tests! :clap:

PR #709, opened 2020-06-10
IE font fix 

danji90:
# How to

The SBB fonts loaded into TrafimageMaps used to use the same name as the font file as the font-family, differing from the SBB standard, potetially causing problems when SBB enbeds Trafimage into their websites. The loadFont function was simplified and now takes an additional argument for the font-family so the name is the same as the SBB standard.

Test the font-family in the netlify [review app](https://deploy-preview-709--dev-trafimage-maps.netlify.app), which should now be 'SBBWeb Roman' instead of 'SBBWeb-Roman' 

# Others

- [x] IE11 tested.
- [x] Tests added.


oterral:
@danji90 please takes time to set up properly the PR.

src/WebComponent.scss
```diff
@@ -1,19 +1,19 @@
-@mixin loadFonts($paths...) {
+@mixin loadFonts($fontFamily, $paths...) {
   @each $family in $paths {
```
oterral, line 2: this loop seems useless now

danji90: tbh, I didn't see the point of the loop before this update either, it's probably been useless for a while. But sometimes I don't question stuff enough. Will try to do that more

oterral:
@danji90  don't forget to set the labels next time hotfix or a release

PR #711, opened 2020-06-17
Map accessibility component

friedjoff:
# How to

Features on Pünktlichkeit layers should be accessible by hitting tab and enter keys.

# Others

- [x] IE11 tested.

### DIFF
--- a/src/components/Map/Map.js
+++ b/src/components/Map/Map.js
@@ -9,6 +9,7 @@ import { unByKey } from 'ol/Observable';
 import OLMap from 'ol/Map';
 import BasicMap from 'react-spatial/components/BasicMap';
 import LayerService from 'react-spatial/LayerService';
+import MapAccessibility from '../MapAccessibility';
 import { setResolution, setCenter, setZoom } from '../../model/map/actions';
 import { setFeatureInfo, setSearchOpen } from '../../model/app/actions';
 
@@ -253,6 +254,7 @@ class Map extends PureComponent {
             extent: maxExtent,
           }}
         />
+        <MapAccessibility layers={layers} map={map} />
       </>
     );
   }
--- a/src/components/Map/Map.js
+++ b/src/components/Map/Map.js
@@ -126,8 +126,8 @@ class Map extends PureComponent {
 
   onPointerDown() {
     const { map } = this.props;
-    if (document.activeElement !== map.getTarget()) {
-      map.getTarget().focus();
+    if (document.activeElement !== map.getTargetElement()) {
+      map.getTargetElement().focus();
     }
   }
 
--- a/src/components/Map/Map.js
+++ b/src/components/Map/Map.js
@@ -74,14 +74,23 @@ class Map extends PureComponent {
 
   componentDidMount() {
     const { map, dispatchHtmlEvent } = this.props;
-    unByKey([this.onPointerMoveRef, this.onSingleClickRef]);
+    unByKey([
+      this.onPointerDownRef,
+      this.onPointerMoveRef,
+      this.onSingleClickRef,
+    ]);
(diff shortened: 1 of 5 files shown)


friedjoff:
> Seems to work partially . I can go through some features but not all of them

Hm, any hint how to reproduce this?

> and we should define an order from left to right and from top to bottom.

Nice idea! Seems a little bit tricky to me, because features are moving. Do you have any idea how to implement this in a reliable way?

FYI: I'm still working on an IE11 bugfix.

oterral:
> Hm, any hint how to reproduce this?

- open https://deploy-preview-711--dev-trafimage-maps.netlify.app/?baselayers=ch.sbb.netzkarte,ch.sbb.netzkarte.luftbild.group,ch.sbb.netzkarte.landeskarte,ch.sbb.netzkarte.landeskarte.grau&lang=de&layers=ch.sbb.puenktlichkeit-all&x=803763.2&y=5905262.21&z=9.713030122207451 
- click on the white portion of the header 
- hit tab 4 times
----> 4 vehciles are selected then cursor goes to the serach

oterral:
> Nice idea! Seems a little bit tricky to me, because features are moving. Do you have any idea how to implement this in a reliable way?

We did something like this in interactive-stationsplans,m  tabs goes from left to right. maybe you can have a look and having some inspiration.

friedjoff:
Did some minor improvements:
- now works in IE11
- select features if map is focused (map can be navigated with arrow keys at the same time)
- sort map features from left to right
- support shift key
- add prop types

Thanks for the bug reproduction hint! I guess this is some kind of race condition: probably the tracker has fetched new trajectories from the server, which haven't been painted to the map yet so they don't have any coordinates. Not sure how to fix this. Maybe we can sell this as a feature (to allow the user to leave the map features)? 😄

Right now it's a little tricky to gain map focus, because we focus the search initally. You will need to use Shift+Tab to find the map focus, but I've added a little visual effect like in IABP. Any idea how to improve this is welcome.

oterral:
>  Maybe we can sell this as a feature (to allow the user to leave the map features)? smile

why not . For me it's ok to be merged. We can fix the bug in another PR but I don't know how right now . Maybe using event on a layer that is triggered when the trajectories are loading or not .

src/components/Map/Map.js
```diff
@@ -73,14 +74,23 @@ class Map extends PureComponent {
 
   componentDidMount() {
     const { map, dispatchHtmlEvent } = this.props;
-    unByKey([this.onPointerMoveRef, this.onSingleClickRef]);
+    unByKey([
+      this.onPointerDownRef,
+      this.onPointerMoveRef,
+      this.onSingleClickRef,
+    ]);
```
oterral, line 81: Since you do this in wiilUnmount. I think it's safe to remove it here.

src/components/Map/Map.js
```diff
@@ -114,6 +124,13 @@ class Map extends PureComponent {
     dispatchHtmlEvent(htmlEvent);
   }
 
+  onPointerDown() {
+    const { map } = this.props;
+    if (document.activeElement !== map.getTarget()) {
+      map.getTarget().focus();
```
oterral, line 130: To be sure you ve got the map element and not a string

```suggestion
      map.getTargetElement().focus();
```

src/components/Map/Map.js
```diff
@@ -114,6 +124,13 @@ class Map extends PureComponent {
     dispatchHtmlEvent(htmlEvent);
   }
 
+  onPointerDown() {
+    const { map } = this.props;
+    if (document.activeElement !== map.getTarget()) {
```
oterral, line 129:
```suggestion
    if (document.activeElement !== map.getTargetElement()) {
```

PR #717, opened 2020-06-25
Improve a11y selection

friedjoff:
# How to

The map is selected after the search field using the tab key now. Also on the Pünktlichkeits-Layer trajectories will be better selectable now.

# Others

- [x] IE11 tested.


oterral:
works better thx

PR #719, opened 2020-06-30
Upgrade react-spatial and use getRenderedTrajectories

friedjoff:
# How to

This PR replaces react-transit with a react-spatial version. Also uses getRenderedTrajectories to improve trajectory a11y selection.

# Others

- [x] IE11 tested.


package.json
```diff
@@ -36,9 +36,8 @@
     "react-redux": "^7.2.0",
     "react-select": "^3.1.0",
     "react-shadow": "^17.6.0",
-    "react-spatial": "^0.3.7",
+    "react-spatial": "0.4.0-beta.1",
```
oterral, line 39: if we merge on master can you use a non-beta release pls?

friedjoff: ✅

PR #743, opened 2020-07-27
Circle markers on start and end of routes only

danji90:
# How to

The route circle markers are currently displayed at the start and end of each route sequence.
These now render only at the start and end coordinates of each route. Unit tests were adjusted accordingly.

Test: https://deploy-preview-743--trafimage-maps.netlify.app

# Others

- [x] IE11 tested.
- [x] Tests added.

### DIFF
--- a/src/layers/RouteLayer/RouteLayer.js
+++ b/src/layers/RouteLayer/RouteLayer.js
@@ -157,9 +157,11 @@ class RouteLayer extends CasaLayer {
             lineStrings.map((l) => l.getGeometry()),
           ),
         });
-        feature.setProperties(sequenceProps);
-        if (isStart) feature.set('start', isStart);
-        if (isEnd) feature.set('end', isEnd);
+        feature.setProperties({
+          ...sequenceProps,
+          isStart,
+          isEnd,
+        });
         return feature;
       });
   }
@@ -177,8 +179,6 @@ class RouteLayer extends CasaLayer {
       this.hoverFeature &&
       (this.hoverFeature.get('route') || {}).isClickable &&
       (this.hoverFeature.get('route') || {}).routeId === routeId;
-    const isStart = feature.get('start');
-    const isEnd = feature.get('end');
 
     const routeStyle = this.styleFunction(
       feature.getProperties(),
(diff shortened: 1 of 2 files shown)


src/layers/RouteLayer/RouteLayer.js
```diff
@@ -192,11 +195,12 @@ class RouteLayer extends CasaLayer {
       const lineStart = feature.getGeometry().getFirstCoordinate();
       const lineEnd = feature.getGeometry().getLastCoordinate();
 
-      styleArray.push(RouteLayer.getCircleStyle(lineStart));
-      styleArray.push(RouteLayer.getCircleStyle(lineEnd));
+      /* Set circle icon if it is the first sequence of a route */
+      if (isStart) styleArray.push(RouteLayer.getCircleStyle(lineStart));
```
mhaertwig, line 199: Use brackets. Putting `isStart` in a variable and renaming it has no effect besides reducing readability.
Directly do `if (feature.get('isStart')`

src/layers/RouteLayer/RouteLayer.js
```diff
@@ -192,11 +195,12 @@ class RouteLayer extends CasaLayer {
...
 
-      /* Set flag icon if it is the final sequence of a route */
-      if (isFinal)
+      /* Set flag and circle icons if it is the final sequence of a route */
+      if (isEnd) {
```
mhaertwig, line 202: `if (feature.get('isEnd'))`

src/layers/RouteLayer/RouteLayer.js
```diff
@@ -157,7 +158,8 @@ class RouteLayer extends CasaLayer {
           ),
         });
         feature.setProperties(sequenceProps);
-        if (finalDestination) feature.set('final', finalDestination);
+        if (isStart) feature.set('start', isStart);
```
mhaertwig, line 161: Always use brackets in if clauses.
Here, you don't need an `if`:

```
feature.setProperties({
  ...sequenceProps,
  isStart, // 1 variable name is enough (not `isStart` and `start`)
  isEnd,
});

danji90: I agree it makes it more readable, but this way we set unnecessary properties for all features (now ALL features receive both properties, isStart and isEnd), inflating the data slightly. But I guess we don't care

src/layers/RouteLayer/RouteLayer.js
danji90: done

src/layers/RouteLayer/RouteLayer.js
danji90: done

PR #746, opened 2020-07-29
Load mbMap with filtered style layers

danji90:
# How to

Updates: 
* Added a filter option to TrafImageMapboxMap
* loadMbMap() overwritten in TrafimageMapboxMap, added filter handling to remove layers from the style object.
* Updated tests

# Others

- [x] IE11 tested.
- [ ] The title means something for a human being.
- [ ] Tests added.

### DIFF
--- a/src/layers/TrafimageMapboxLayer/TrafimageMapboxLayer.js
+++ b/src/layers/TrafimageMapboxLayer/TrafimageMapboxLayer.js
@@ -108,10 +108,15 @@ class TrafimageMapboxLayer extends MapboxLayer {
     }
 
     /* Get the MapBox style */
-    const mapboxStyle = await getMapboxStyle(this.styleUrl);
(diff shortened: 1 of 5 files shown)


src/layers/TrafimageMapboxLayer/TrafimageMapboxLayer.js
```diff
@@ -70,6 +95,114 @@ class TrafimageMapboxLayer extends MapboxLayer {
...
+         * Copyright statement.
+         * @type {string}
+         */
+        this.copyright = getCopyrightFromSources(this.mbMap);
+      }
+      this.dispatchEvent({
+        type: 'load',
+        target: this,
+      });
+      this.olLayer.changed();
```
oterral, line 191: maybe write why we need to do this

PR #747, opened 2020-07-30
Hotfix the search clear button

edellucien:
# How to

Clear button for the search bar is hidden behind the search button.
Ticket: TRAFWART-858

# Others

- [x] IE11 tested.
- [ ] Tests added.

### DIFF
--- a/src/components/Search/Search.scss
+++ b/src/components/Search/Search.scss
@@ -32,6 +32,7 @@
     &.wkp-search-button-clear {
       background: none;
       color: #666;
+      right: 48px;
     }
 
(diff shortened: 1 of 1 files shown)


src/components/Search/Search.scss
```diff
@@ -32,6 +32,7 @@
     &.wkp-search-button-clear {
       background: none;
       color: #666;
+      right: 48px;
```
oterral, line 35: @edellucien  the buttons clear and search are inside   flex container for the positionning .  I think you can remove the absoulte,right,top  positionning  instead of adding this css. But maybe I miss something

PR #757, opened 2020-08-14
WIP: Allow to pass a menubottom component to topics

edellucien:
# How to

1/ Change Webcomponents.scss to be able to Load WebComponents directly in trafimage-maps on gitlab, and therefore avoid the problem of two react versions loaded.
2/ Possiblity to pass a topicMenuBottom component (for immobilien filters)

# Others

- [ ] IE11 tested.
- [ ] Tests added.


src/components/TrafimageMaps/TrafimageMaps.js
```diff
@@ -160,6 +161,28 @@ class TrafimageMaps extends React.PureComponent {
      * @private
      */
     this.store = getStore();
+
+    this.theme = createMuiTheme({
```
friedjoff, line 165: Is this required for the optional `TopicMenuBottom` component? If so I would prefer wrapping just `TopicMenuBottom` in `TopicMenu` with the provider or at least have a comment here. Otherwise we'll forget what this is about.

friedjoff:
@edellucien can you also fix the CI?

PR #795, opened 2020-09-08
Add apiKeyName api parameter

gevos:
See ROKAS-68

# How to

The application should behave the same. All api requests should use the api key 'key'. However, it's possible to globally change the api key name.

Test WKP with modifications
- Open the Network tab in the developer tools
- Run https://github.com/geops/trafimage-maps/pull/813 locally
=> The basemap does not load
- Search for 'api_key' in the Network tab of the developer tools
=> `https://maps.geops.io/styles/base_bright_v2/style.json...` should be listed with an apiKey `12345`
- Activate layer "Pünktlichkeit"
- Search for anything
=> The parameter `key` (not `api_key`) should be used with a apiKey `12345` for
`..trajectory_collection?...` 
`https://api.geops.io/stops/v1/?&q=Ber&key=12345`
=> Additionally `Unhandled Rejection (TypeError): Cannot read property 'slice' of undefined` appears as no items exist

Test WKP without modifications
- Use the test app from https://github.com/geops/trafimage-maps/pull/814
=> All layers/services should work

Test Casa Map with modifications
- Open the Casa Map example in the test app from https://github.com/geops/trafimage-maps/pull/822
- Open the Network tab in the developer tools and search for 'api_key'
=> The urls should be 
`https://api.geops.io/routing/v1/TEST/`
`https://maps.geops.io/styles/base_bright_v2/style.json`

Test Casa Map without modifications
- Use the test app for Casa Map from https://github.com/geops/trafimage-maps/pull/814
=> All layers should work

# Others

- [x] Everything in ticket description has been fixed. (Together with https://github.com/geops/trafimage-maps/pull/797)
- [ ] IE11 tested.
- [ ] Tests added.


src/components/TopicLoader/TopicLoader.js
```diff
@@ -120,7 +123,9 @@ class TopicLoader extends Component {
 
     if (
       activeTopic &&
-      (language !== prevProps.language || apiKey !== prevProps.apiKey)
+      (language !== prevProps.language ||
+        apiKey !== prevProps.apiKey ||
+        apiKeyName !== prevProps.apiKeyName)
```
gevos, line 128: Can the key and its key name really change?

src/components/TopicLoader/TopicLoader.js
```diff
@@ -294,7 +301,7 @@ class TopicLoader extends Component {
         flatLayers[i].getApiKey &&
         !flatLayers[i].getApiKey()
       ) {
-        flatLayers[i].setApiKey(apiKey);
+        flatLayers[i].setApiKey(apiKey, apiKeyName);
```
gevos, line 304: As far as I know, this is currently only relevant for react-transit TrajservLayer (see https://github.com/geops/react-transit/blob/master/src/layers/TrajservLayer.js#L484). I will ask in the ticket, if react-transit changes are necessary. Even if not, I would pass the apiKeyName here for future layers and consistency.

src/layers/RouteLayer/RouteLayer.js
```diff
@@ -137,7 +137,7 @@ class RouteLayer extends CasaLayer {
 
     const via = viaPoints.map((v) => `!${v}`);
     const urlParams = {
-      key: this.apiKey || '',
+      [this.apiKeyName || 'key']: this.apiKey || '',
```
gevos, line 140: Alternatively, we could defined the default value in the CasaLayer. Same for ZoneLayer.

src/components/TopicLoader/TopicLoader.js
edellucien: Well I think so if we wait on a fetchApiKey call, no? It changes from undefined to something

edellucien (approved):

src/components/TopicLoader/TopicLoader.js
gevos: Done.

gevos:
I have removed the apiKeyName change in the mobility-toolbox-js for the TrajservLayer.

edellucien:
Thanks for all the test branches. ALl of them work well. :+1:

PR #802, opened 2020-09-14
Construction: Move layer and add info texts

gevos:
# How to

See TRAFMBB-20

- The layer "Angebotsschritt 2035" was moved as sublayer to "Ausbau". Toggling all sublayers of "Ausbau" should still work. If it's a sublayer which contributes to the clustering (non-"Angebotsschritt 2035"), the clustering must be updated. Otherwise, nothing should happen.
- Use the same info text for "Angebotsschritt 2035" as for its sublayer with the date 2035.

# Others

- [ ] IE11 tested.


friedjoff (approved):
LGTM

PR #804, opened 2020-09-16
Passenger frequencies - Circle area instead of circle radius

danji90:
# How to

**Task**: Passenger frequency numbers should be proportional to the circle area instead of the circle radius

**Solution**: Mapbox circle paint properties have no circle-area property, only circle-radius. Hence the square root of the passenger frequency is used to calculate the radius, instead of the absolute values, producing the desired result.

**Test**: Compare passenger frequencies layer in the [review app](url) the prod version and see the difference

# Others

- [x] IE11 tested.
- [ ] Tests added.


friedjoff (approved):
Nice!

PR #809, opened 2020-09-28
Encode search queries

gevos:
...for:

- StopFinder
- Betriebspunkte
- Lines
- Locations
- Municipalities

...which fixes the search in IE with umlauts.

See TRAFMBB-21

Same as https://github.com/geops/react-transit/pull/285 from https://jira.geops.ch/browse/MAPSET-59

# How to

- Open review app in IE
- Search for "Zür"
=> The Stationen "Zürich HB" and "Zürich Oerlikon" should appear.

# Others

- [ ] IE11 tested.
- [ ] Tests added.

### DIFF
--- a/src/searches/Betriebspunkte/Betriebspunkte.js
+++ b/src/searches/Betriebspunkte/Betriebspunkte.js
@@ -11,7 +11,9 @@ class Betriebspunkte extends Search {
 
   // eslint-disable-next-line class-methods-use-this
   search(value) {
-    return fetch(`https://maps.trafimage.ch/search/bps?name=${value}`)
+    return fetch(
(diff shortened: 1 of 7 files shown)


edellucien:
When I search with 'Zür' I actually get results for Stationen, but not for further categories, such as 'Linien', 'Gemeinden', 'Orte', etc.. for topic Netzkarte.
We should encode by default for all searches, do you agree?

edellucien (changes requested):
When I search with 'Zür' I actually get results for Stationen, but not for further categories, such as 'Linien', 'Gemeinden', 'Orte', etc.. for topic Netzkarte.
We should encode by default for all searches, do you agree?

edellucien:
It works correctly now. Thanks for the fix :+1:

PR #812, opened 2020-09-29
Fix menu which was overlapped by the search at large (.tm-w-l) screens

gevos:
Bug without ticket

The element .wkp-search must not overlap the menu.

# How to

- Open the application at a screen width of .tm-w-l
- Set the cursor into the search field
- Try to click on the arrow of the menu
=> The menu should open

The same should happen for all screen widths. 

# Others

- [ ] Everything in ticket description has been fixed. - There is no ticket
- [ ] IE11 tested.
- [ ] Tests added.


edellucien:
Good catch thanks!

PR #826, opened 2020-10-21
Allow handicap popup content to be read by screen reader

edellucien:
# How to

When using a screenreader, the handicap popup content must be read out loud when opened.
Solution inspired from docu:
https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html
Ticket: TRAFWART-910

Tested with ORCA on linux & nvaccess on windows

# Others

- [x] IE11 tested.
- [ ] Tests added.

### DIFF
--- a/src/popups/HandicapPopup/HandicapPopup.js
+++ b/src/popups/HandicapPopup/HandicapPopup.js
@@ -14,15 +14,12 @@ function HandicapPopup({ feature }) {
   const properties = feature.getProperties();
   const { t } = useTranslation();
   const refBody = useRef();
-  const refBodyEmpty = useRef();
 
   useEffect(() => {
     // focus first element to trigger screenreader to read its content:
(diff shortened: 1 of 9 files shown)


src/popups/HandicapPopup/HandicapPopup.js
```diff
@@ -174,11 +186,17 @@ function HandicapPopup({ feature }) {
 
   const renderBody = () => {
     if (properties.noInfo) {
-      return <span>{t('Keine Information vorhanden.')}</span>;
+      return (
+        <span tabIndex={-1} ref={refBodyEmpty}>
```
oterral, line 190: you can use refBody

edellucien: Thanks for the advice, done as so.

oterral (approved):
lgtm

PR #859, opened 2021-01-05
Define unique keys to showcase layers

edellucien:
# How to
Define unique keys to showcase layers

# Others

- [ ] IE11 tested.
- [ ] Tests added.

### DIFF
--- a/src/config/layers.js
+++ b/src/config/layers.js
@@ -508,6 +508,7 @@ export const nichtBarrierfreierBahnhoefe = new MapboxStyleLayer({
 
 export const netzkarteShowcasesNight = new TrafimageMapboxLayer({
   name: 'ch.sbb.netzkarte.night',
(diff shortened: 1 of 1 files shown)


src/config/layers.js
```diff
@@ -534,7 +536,7 @@ export const netzkarteShowcasesLight = new TrafimageMapboxLayer({
 
 export const netzkarteShowcasesNetzkarte = new TrafimageMapboxLayer({
   name: 'ch.sbb.netzkarte.layer',
-  key: 'ch.sbb.netzkarte',
+  key: 'ch.sbb.netzkarte.layer',
```
mhaertwig, line 539: Why `.layer`? Old permalinks could stop working.

edellucien: For consistency, to have similar key and name, but we can keep the key without '.layer' sure :)

PR #872, opened 2021-02-17
Fix CASA examples

danji90:
# How to

The filters for the CASA labels layer (overlayers all other layers) was no longer functional, since the filter application was removed in this [commit|https://github.com/geops/trafimage-maps/commit/6f19af92763729b5c4438e9419bec17670019308] (it runs into both of the conditions that return (abort) in setStyleConfig. This caused the labels layer to render all layers, completely overlaying everything else. 

The condition was reinserted to force apply of the filter.

# Others

- [ ] IE11 tested.
- [ ] Tests added.


oterral:
Fixed by https://github.com/geops/trafimage-maps/pull/873

PR #897, opened 2021-04-19
Parse tourist offer data string in popup

danji90:
# How to

Publishing Cartaro vector tile data only allows serialization into types String, Number or Boolean, resulting in array data as a String. We therefore need to parse the string in the frontend 

# Others

- [ ] Tests added.


danji90:
@oterral, I got the LGTM from @mhaertwig

PR #899, opened 2021-04-29
Disable pagination for zweitausbildung poi popup.

mhaertwig:
Introduce `hidePagination` popup property, display all popup items for zweitausbildung pois without pagination.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "trafimage-maps",
-  "version": "v1.5.14-beta.3",
+  "version": "1.5.14-beta.4",
   "private": true,
   "main": "build/bundle.js",
   "proxy": "http://127.0.0.1:8000",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "trafimage-maps",
-  "version": "1.5.11",
+  "version": "1.5.13",
```
mhaertwig, line 3: Didn't count for a long time.

oterral (changes requested):
With this PR you fix the issue but you potentially creates others. For example if there is another layer that is queryable , the pagination will not be displayed. 

The Popup should not hide the pagination but  should decide if  features of the corresponding layers are displayed in one page or one pro page. and if it's the case the `features` array should not be a flat map of features but an array containing feature or array of features.

```
const features = [
feature1Layer1,
feature2Layer1,
[ feature1Layer2, feature2Layer2 ]
]; 
```

like this you keep the pagination and you avoid adding a new  property to the popup.  Or even better you place the `feature` property by  the  `features` property.

oterral (approved):

oterral:
I merge this since the topic is already on production

PR #911, opened 2021-07-06
Copyright as standalone component

danji90:
# How to

Problem: Currently the copyright is coupled with the app footer, resulting in a missing copyright when loading the web component without the footer. CASA does not want the footer, but needs the copyright.

Solution: The Footer was moved into a standalone component entirely. It was added to the wkp default elements, to CASA default elements (CSS also adjusted for CASA in casa.css) and to the style guide examples as needed.

# Others

- [ ] Tests added.

### DIFF
--- a/src/components/Copyright/Copyright.js
+++ b/src/components/Copyright/Copyright.js
@@ -4,7 +4,7 @@ import { useSelector } from 'react-redux';
 import RsCopyright from 'react-spatial/components/Copyright';
 import { makeStyles } from '@material-ui/core';
 
-const useStyles = makeStyles((theme) => ({
+const useStyles = makeStyles(() => ({
   wrapper: {
     position: 'absolute',
     right: 5,
@@ -14,10 +14,6 @@ const useStyles = makeStyles((theme) => ({
     '& a:not(.MuiIconButton-root)': {
       whiteSpace: 'nowrap',
       textDecoration: 'none !important',
-    },
-    [theme.breakpoints.down('md')]: {
-      fontSize: 11,
-      right: 'unset',
     },
   },
 }));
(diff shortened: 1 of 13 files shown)


src/components/Footer/Footer.js
```diff
@@ -3,7 +3,7 @@ import { useTranslation } from 'react-i18next';
 import { useDispatch, useSelector } from 'react-redux';
 import UIFooter from '@geops/react-ui/components/Footer';
 import ScaleLine from 'react-spatial/components/ScaleLine';
-import Copyright from 'react-spatial/components/Copyright';
+// import RsCopyright from 'react-spatial/components/Copyright';
```
mhaertwig, line 6: This can be removed.

src/components/Copyright/Copyright.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { useSelector } from 'react-redux';
+import RsCopyright from 'react-spatial/components/Copyright';
+import { makeStyles } from '@material-ui/core';
+
+const useStyles = makeStyles((theme) => ({
```
mhaertwig, line 7: Nice you used MUI styling 👍

src/components/TopicElements/TopicElements.js
```diff
@@ -43,6 +44,7 @@ const defaultElements = {
   exportMenu: false,
   search: false,
   overlay: false,
+  copyright: true,
```
mhaertwig, line 47: The copyright needs to be always visible and should not be configurable.

mhaertwig (changes requested):
Nice. The most important change is that the copyright should not be configurable.

src/components/Copyright/Copyright.js
danji90: I actually just noticed the MUI style solution is not ideal in this case, since the MUI breakpoints always refer to the window width, regardless how big the web component container is :/ Will find an alternate solution

PR #923, opened 2021-08-18
Fix lines Search

danji90:
# How to

The Search for line kilometre points is not working and the Search line segments works but doesn't highlight the line on the map.

Fix: 
- Updated the faulty regex for kilometre search (valid searches for line 500 and km 50.1: 500 50.1, 500 +50.1, 500 + 50.1)
- Added Point highlighting when Point is returned (e.g. a kilometer)
- Fixed line highlighting on map by accepting GeometryCollections (as well as MultiLineString)

Test: 
- **500 50.1**, **500 +50.1**, **500 + 50.1** should return a point in lines search results and highlight it on the map on hover
- **500 1-80** should return a line segment in the lines results and highlight the line on hover  

# Others

- [x] Tests added.

### DIFF
--- a/src/searches/Lines/Lines.js
+++ b/src/searches/Lines/Lines.js
@@ -1,6 +1,11 @@
 import React from 'react';
 import { Style, Circle, Fill, Stroke as OLStroke } from 'ol/style';
-import { MultiLineString, GeometryCollection } from 'ol/geom';
+import {
+  MultiLineString,
+  GeometryCollection,
+  MultiPoint,
+  Point,
+} from 'ol/geom';
 import GeoJSON from 'ol/format/GeoJSON';
 
 import Search from '../Search';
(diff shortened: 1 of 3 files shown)


src/searches/Lines/Lines.js
```diff
@@ -10,7 +15,9 @@ const lineMeasuresRegExp = new RegExp(
   '([0-9]*)\\s*([0-9]+\\.?[0-9]*)\\-([0-9]*\\.?[0-9]*)',
 );
 
-const lineKilometerRegExp = new RegExp('^([0-9]+)\\s+([0-9]+\\.?[0-9]+)$');
+const lineKilometerRegExp = new RegExp(
+  '^([0-9]+)(\\s)(\\+)?(\\s+)?([0-9]+(\\.[0-9]+)?)$',
+);
```
oterral, line 20: I think there is  test somewhere for this regex, can you check and if not can you add tests for the new possible values.pls

oterral (approved):
lgtm . Just add test for the regex.

src/searches/Lines/Lines.js
danji90: I didn't find any present tests, so I added some in Lines.test.js

PR #924, opened 2021-08-18
Block Popups parameter

danji90:
# How to

A permalink parameter needs to be added that blocks all popups

# Others

- [ ] Tests added.


oterral (changes requested):
There is still the mouseover effect (test passagierfrequenzen layer for example). Should we remove the complete effect on pointermove as well ?

PR #946, opened 2021-11-18
Beleuchtungsstärken topic

danji90:
A new topic "Beleuchtungsstärken" needs to be added to trafimage-maps. It shows the illuminance levels at swiss train stations. The topic data is not managed by Cartaro, but rather imported annually with the updated data.

# How to
The topic contains a togglable layer for each illuminance level. A layer popup and a topic info popup shows the legend with information about the illuminance levels.

Some extra-Layers showing geo-data from the FOEN have also been added, to be reviewed by the client
# Others

- [ ] Tests added.


oterral:
lgtm. I wait to see with some icons on the map.

danji90:
> lgtm. I wait to see with some icons on the map.

Dunno why, but the maps.geops.io takes ages to fully deploy. Now all the data is here.
There are still some features missing though which we need to add before we merge, so we leave it as a draft for now

PR #952, opened 2021-12-06
Deactivate completely pointer events when popup element is disabled

oterral:
# How to

Open an example of the doc and set the topic.elements.popup property to false. There should not be highlight and select effect.

# Others

- [x] Tests added.

### DIFF
--- a/src/components/TopicElements/TopicElements.js
+++ b/src/components/TopicElements/TopicElements.js
@@ -74,8 +74,12 @@ const getComponents = (defaultComponents, elementsToDisplay) =>
 function TopicElements({ history, appBaseUrl, loginUrl, staticFilesUrl }) {
   const ref = useRef(null);
   const dispatch = useDispatch();
-  const { activeTopic, layerService, map } = useSelector((state) => state.app);
+  const { t } = useTranslation();
+  const activeTopic = useSelector((state) => state.app.activeTopic);
+  const layerService = useSelector((state) => state.app.layerService);
+  const map = useSelector((state) => state.app.map);
   const [tabFocus, setTabFocus] = useState(false);
+
   useEffect(() => {
     const unfocusTab = () => setTabFocus(false);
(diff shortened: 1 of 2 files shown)


danji90:
Works in the casa example, so I guess it works in general. Should the tracker popup (RouteSchedule) also be affected?
BTW, the Construction example points to _Netzkarte_ but should be pointing at _Bauprojekte Bahninfrastruktur_ right?

src/components/TopicElements/TopicElements.js
```diff
@@ -74,8 +74,12 @@ const getComponents = (defaultComponents, elementsToDisplay) =>
 function TopicElements({ history, appBaseUrl, loginUrl, staticFilesUrl }) {
   const ref = useRef(null);
   const dispatch = useDispatch();
-  const { activeTopic, layerService, map } = useSelector((state) => state.app);
+  const { t } = useTranslation();
+  const activeTopic = useSelector((state) => state.app.activeTopic);
+  const layerService = useSelector((state) => state.app.layerService);
+  const map = useSelector((state) => state.app.map);
```
danji90, line 80: Is this for readability or performance? Just out of interest...

danji90 (commented):
The code and functionality looks good, LGTM after double checking the two questions/comments I left

src/components/TopicElements/TopicElements.js
oterral: Question of performance. It's bad practice to listen the all state.app object because it will rerender the component  every time the state change, not only when actievTopic, layerService amd map change.
 So listnening each property separetely will rerender the component only when the property change.

oterral:
yes it should netzkarte but another PR already changes that.

oterral:
> Should the tracker popup (RouteSchedule) also be affected?

Since the routeSchedule  is configured with the elements.trackerMenu property  i let it as it was.

PR #963, opened 2022-01-05
Fix app interaction while Dialog is open

danji90:
# How to

The new MUI Dialog (LayerInfo, TopicInfo) has some conflicts with certain DOM Elements while the Dialog is visible.
- Select Dropdowns (Language, Projection) can be clicked to open and hover styles apply, but the menu can't be toggled anymore, doesn't close and click on a menu option has no effect.

Since we are switching many components to MUI components and we already have an SBBSelect, the dropdowns can be solved by using the SBBSelect for all DD-Menus

# Others

- [ ] Tests added.


danji90:
@oterral _disableEnforceFocus: true_ seems to have done the trick

oterral (changes requested):
It seems there is duplicated styles in langSelect and projectionSelect, can we put them in the generic Select component ?

oterral (changes requested):
- margin between icon and text are not the same for login text and language box
- language box is not aligned with login text

danji90:
@oterral LGTM?

src/components/Select/Select.js
```diff
@@ -1,67 +1,83 @@
...
+  return {
+    outlineHidden: {
+      '&:hover': {
+        color: theme.palette.secondary.dark,
+      },
+      '&:hover .MuiOutlinedInput-notchedOutline,& .MuiOutlinedInput-notchedOutline':
+        {
+          borderWidth: 0,
+        },
+      '& Mui-focused .MuiOutlinedInput-notchedOutline': {
```
oterral, line 16: & .Mui-focused

oterral, line 16:
```suggestion
      '& .Mui-focused .MuiOutlinedInput-notchedOutline': {
```

oterral:
@danji90  there is BAFU layers change in this PR is it normal ?

oterral (approved):
Very nice thanks, Ihave fixed the border of the select box on mobile

PR #967, opened 2022-01-12
Improve Select

danji90:
# How to

Updates:
- Fixed the hideOutline MUI error message by deleting the prop before injecting them
- Using deepMerge to merge the MenuProps. MenuProps injected into the local Select component are deep merged with a default MenuProps object inside the component. This saves us from applying a lot of code when calling the component.
- Using TransintionProps _onEnter_ (as suggested by MUI, since PaperProps _onEnter_ is deprecated). By using this property in the default MenuProps object the dropdown width is now always calculated automatically

# Others

- [x] Tests added.


oterral (approved):
Very nice thx

PR #974, opened 2022-01-31
Add default sbb consent management

oterral:
# How to

Disply the official SBB consent window if enableTracking is true.

# Others

- [x] Tests added.


cypress/integration/Search.spec.js
```diff
@@ -5,6 +5,7 @@ describe('Search', () => {
     cy.visit('');
   });
   it(`should display plural texts`, () => {
+    cy.get('#onetrust-accept-btn-handler').click();
```
friedjoff, line 8: Maybe move into `beforeEach`?

src/WebComponent.js
```diff
@@ -133,22 +133,12 @@ const propTypes = {
   /**
    * Enable analytics tracking.
    */
-  enableTracking: PropTypes.bool,
+  enableTracking: PropTypes.string,
```
friedjoff, line 136: String because of web component?

src/components/Head/Head.js
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import { useSelector } from 'react-redux';
+import PropTypes from 'prop-types';
+import Helmet from 'react-helmet';
+
+/**
+ * This component add tag in the htm head tag for sbb consent management.
```
friedjoff, line 7:
```suggestion
 * This component adds a script tag in the HTML head for SBB consent management.
```

src/components/MatomoTracker/MatomoTracker.js
```diff
@@ -0,0 +1,64 @@
...
+  useEffect(() => {
+    if (!activeTopicKey || !consentGiven || !trackEvent || !localStorage) {
+      return;
+    }
+    const startDate = localStorage?.getItem(LS_MATOMO_USER_SESSION_TIMER) || 0;
+    const topicVisited = localStorage?.getItem(LS_MATOMO_TOPIC_VISITED) || '';
+
+    // A matomo user session is 30 min by default.
+    if (
+      Date.now() - startDate > 30 * 60 * 1000 ||
```
friedjoff, line 45: Maybe extract `30 * 60 * 1000` into a `const` variable?

friedjoff (commented):
Code looks solid! Just small suggestions...

src/WebComponent.js
oterral: yes when you specify an attribute in the webcomponent it has to be string. A property can be somthing else but not an attribute.

cypress/integration/Search.spec.js
oterral: Done

PR #1016, opened 2022-07-19
Draft: Use base_bright_v2_legacy in Netzkarte Sandbox

danji90:
# How to
Currently Netzkarte Sandbox uses _base_bright_v2_ch.sbb.netzkarte_, loading a lot of data the topic doesn't need (e.g. Direktverbindungen). Instead, we could use the base style base_bright_v2 or create a new topic-specific partial.

# Others

- [ ] Tests added.


oterral:
It has been decide to fix the netzkarte partial. There was a problem with an highlight style which was visible and load the tile from direktverbindung

PR #1037, opened 2022-10-06
fix: display correct subtitle in menu header

danji90:
# How to
Since the LayerrService was migrated into Trafimage and the change:visible listeners were commented out, the TopicMenuHeader can no longer listen to this event and the subtitle does not update

# Others

- [x] Tests added.


oterral:
Indeed it was the poroblem thx for the fix.

At the end the layerService will be removed and replaced by utils function.

oterral (approved):

PR #1053, opened 2022-12-14
fix(Tarifverbundkarte): improved popup and compile legend using style metadata

danji90:
# How to

# Others

- [x] Tests added.


.env
```diff
@@ -4,7 +4,7 @@ REACT_APP_CARTARO_URL=//cartaro2.trafimage.ch/api/v1/
 REACT_APP_BASE_URL=//maps.trafimage.ch
 REACT_APP_MATOMO_URL_BASE=https://analytics.geops.de/
 REACT_APP_MATOMO_SITE_ID=9
-REACT_APP_VECTOR_TILES_URL=https://maps.geops.io
+REACT_APP_VECTOR_TILES_URL=https://maps.style-dev.geops.io
```
danji90, line 7: This needs to be reversed before merge

src/layerInfos/TarifverbundkarteTopicInfo/TarifVerbundLegend.js
```diff
@@ -0,0 +1,68 @@
...
+    fetch(`${vectorTilesUrl}/styles/ch.sbb.tarifverbund/style.json`)
+      .then((response) => response.json())
+      .then((data) =>
+        setVerbunde(
+          data.metadata.partners.sort((a, b) => a.name.localeCompare(b.name)),
+        ),
+      )
+      // eslint-disable-next-line no-console
+      .catch((err) => console.error(err));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
oterral, line 30: can you cancel the request when necessary  ?

oterral: maybe return an empty array when the request fails

oterral (changes requested):
Tests are missing otherwise lokks good to me . Also make sure that the app doesn't break if the request fails

src/layerInfos/TarifverbundkarteTopicInfo/TarifVerbundLegend.js
```diff
@@ -0,0 +1,68 @@
...
+        props.isMobile ? '1fr 1fr' : '200px 200px',
+      gap: 5,
+    },
+  };
+});
+
+const useFetchVerbunde = (vectorTilesUrl) => {
+  const [verbunde, setVerbunde] = useState([]);
+  useEffect(() => {
+    fetch(`${vectorTilesUrl}/styles/ch.sbb.tarifverbund/style.json`)
```
oterral, line 21: Are you sure we need this fetch you already have the style using the mapbox map directkly(accessing via the layer)

src/layerInfos/TarifverbundkarteTopicInfo/TarifVerbundLegend.js
danji90: The defaultIt already returns an empty array, since the default state in useState is set to an empty array

src/layerInfos/TarifverbundkarteTopicInfo/TarifVerbundLegend.js
danji90: No that won't work I think, because the User can view the Tarifverbundkarte LayerInfo even when the Topic is not active. If the Topic is not active, the Layer mbMap tag is undefined

danji90:
@oterral I added a test for the legend and made sure the request fails are handled

PR #1073, opened 2023-03-13
fix: use one websocket for all punctuality layers and use the proper …

oterral:
…api key for the websocket

# How to

Display punctuality layers , open network tab ---> only one websocket is open and use the api key defined

# Others

- [x] Tests added.


src/config/ch.sbb.netzkarte/index.js
```diff
@@ -184,111 +184,94 @@ export const punctuality = new Layer({
...
+      properties: {
+        filter: ({ properties }) => {
+          const { type, line } = properties;
+          return type === 'rail' && /(IR|IC|EC|RJX|TGV)/.test(line?.name);
+        },
+      },
+    }),
+  ];
+  punctuality.children.forEach((layer) => {
+    layer.on('change:visible', () => {
```
danji90, line 249: @oterral I'm pretty sure it's a bug in the review app only (I've seen this before), but turning on the child layers does not activate the  parent layers (not only in punctuality), so if punctuality is turned off and then you turn on an mot it doesn't open the web socket connection because TralisLayer visibility is not updated. Double check that it works when using it in WKP. Otherwise LGTM

danji90 (approved):
Plz double check functionality when using as npm package, I can't fully verify it works in the review app (see comment). Otherwise LGTM

danji90:
@oterral If you have time, the cypress tests need to be fixed. I think there's a caching problem or something

PR #1131, opened 2024-01-16
fix(map controls): move map controls with overlay so they are always visible

danji90:
# How to
MapControls refactoring:
- Remove MapControls scss, use MUI jss instead
- Made all controls MapButton components (except for Zoom, which comes from react-spatial)
- Save overlay element in state so it can be used across the app
- Use the overlay element width to relocate MapControls component

# Others

- [ ] Tests added.


src/components/MapControls/MapControls.js
```diff
@@ -40,13 +43,102 @@ const degreesToRadians = (degrees) => {
...
 };
 
+const useStyles = makeStyles((theme) => {
+  return {
+    mapControls: {
+      position: "absolute",
+      display: "flex",
+      flexDirection: "column",
+      gap: 8,
+      right: (props) => (props.isMobile ? 15 : props.overlayWidth + 15),
```
oterral, line 53: it should be 10 instead of 15 on small width screen

src/components/MapControls/MapControls.js
```diff
@@ -40,13 +43,102 @@ const degreesToRadians = (degrees) => {
   return degrees * (Math.PI / 180);
 };
 
+const useStyles = makeStyles((theme) => {
+  return {
+    mapControls: {
+      position: "absolute",
+      display: "flex",
+      flexDirection: "column",
+      gap: 8,
```
oterral, line 52: should be also 10 on mobile and 15 on desktop

src/components/MapControls/MapControls.js
```diff
@@ -40,13 +43,102 @@ const degreesToRadians = (degrees) => {
...
+            cursor: "default",
+          },
+        },
+      },
+    },
+    "@keyframes fadeInOut": {
+      "0%": { opacity: 0 },
+      "50%": { opacity: 1 },
+      "100%": { opacity: 0 },
+    },
```
oterral, line 92: put keyframes definition  in globals.scss so we can reuse it

src/components/MapControls/MapControls.js
```diff
@@ -40,13 +43,102 @@ const degreesToRadians = (degrees) => {
...
+    geolocation: {
+      padding: 0,
+      color: theme.colors.darkGray,
+      "&:hover": {
+        color: theme.colors.darkGray,
+      },
```
oterral, line 98: can this color be defined in MapButton ? it seems redundant in fitExtent as well

src/utils/useOverlayWidth.js
```diff
@@ -0,0 +1,21 @@
...
+
+const useOverlayWidth = () => {
+  const overlayElement = useSelector((state) => state.app.overlayElement);
+  const [width, setWidth] = useState(0);
+  useEffect(() => {
+    const resizeObserver = new ResizeObserver((entries) => {
+      setWidth(entries[0].contentRect.width);
+    });
+    if (!overlayElement) {
+      resizeObserver.disconnect();
```
oterral, line 12: disconnect in the function returned by the use effect

src/utils/useOverlayWidth.js
```diff
@@ -0,0 +1,21 @@
...
+      setWidth(0);
+      return;
+    }
+    resizeObserver.observe(overlayElement);
+  }, [overlayElement]);
```
oterral, line 17: always return a function when you unlisten the resize

src/components/Overlay/Overlay.js
```diff
@@ -240,6 +244,9 @@ function Overlay({
             return null;
           },
         }}
+        PaperProps={{
+          ref: (node) => dispatch(setOverlayElement(isMobile ? null : node)),
```
oterral, line 248: To make it more understandable with the useEffect at the beginning do:

Here do: 
ref => (node) => setOverlayNode(node)

then put the logic in the useEffect():
useEffect(() => {
    dispatch(setOverlayElement(isMobile ? null : node))
    return () => dispatch(setOverlayElement(null));
  }, [isMobile, node, dispatch]);

oterral (approved):
If you choose to remove the scss  file, make sure you are not breaking any css  for other screen size. And remove the scss clas s not used  or useless.

If you have time maybe it could be nice to extract Geolocation and MapExtent in their own component it will increase the readibility.

src/components/MapControls/MapControls.js
danji90: yeah makes sense

src/utils/useOverlayWidth.js
danji90: Done

src/utils/useOverlayWidth.js
danji90: Done

src/components/Overlay/Overlay.js
danji90: Done

src/components/MapControls/MapControls.js
danji90: ok

src/components/MapControls/MapControls.js
danji90: OK

PR #1136, opened 2024-01-23
fix: change angular example

oterral:
# How to

Fix the angular example and change the way to load the base layers pngs so they are available in the web components

# Others

- [x] Tests added.


src/components/TopicElements/TopicElements.js
```diff
@@ -106,6 +108,18 @@ function TopicElements({ history }) {
     );
   }, [activeTopic]);
 
+  // We define the images for the base layers using staticFilesUrl because they are too big to be encoded in base64.
+  // so when using in the web component (see angular example) the app is able to get them.
+  // Images are stored in the gitlab project.
+  const baseLayerImages = useMemo(() => {
+    const imgs = {};
+    baseLayers.forEach((layer) => {
+      imgs[layer.key] =
```
danji90, line 117: This doesn't work, the images are not found in the Angular example because the bundle renames the static images adding a random string:
https://maps.trafimage.ch/static/media/netzkarte_night.2ba07b051eec597a85cb.png

Also, in the Angular example the wrong static files url is formed, the requests for the images are made like this:
https://4jkf9x-4200.csb.app/static/media/luftbild.64e603fa2f0d815af1f0.png

danji90 (changes requested):
@oterral in my tests, it seems to me like this approach doesn't work. Did you test?

src/components/TopicElements/TopicElements.js
oterral: it works as expected it is just that the angular example uses the last published version not the one in htis PR .

I have tested on https://dev-trafimage-maps-git-olivier-angular-geops.vercel.app/?lang=de&x=928460&y=5908948&z=8.5&layers=&baselayers=ch.sbb.netzkarte%2Cch.sbb.netzkarte.dark%2Cch.sbb.netzkarte.luftbild.group%2Cch.sbb.netzkarte.landeskarte%2Cch.sbb.netzkarte.landeskarte.grau

PR #1194, opened 2024-10-09
fix(Regionenkarte): prevent line search highlight from disappearing onPointerMove

danji90:
# How to

# Others

- [ ] Tests added.

### DIFF
--- a/src/config/ch.sbb.regionenkarte.public/index.js
+++ b/src/config/ch.sbb.regionenkarte.public/index.js
@@ -65,7 +65,6 @@ export const regionenkartePublicSegment = new Layer({
       },
       properties: {
         hideInLegend: true,
-        showPopupOnHover: true,
         popupComponent: "RegionenkarteIntersectionPopup",
         isQueryable: true,
       },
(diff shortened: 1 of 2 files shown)


src/config/ch.sbb.regionenkarte.public/index.js
```diff
@@ -65,7 +65,6 @@ export const regionenkartePublicSegment = new Layer({
       },
       properties: {
         hideInLegend: true,
-        showPopupOnHover: true,
```
danji90, line 68: @oterral We don't need this here anymore right? It was preventing the select highlight from lingering

danji90: And do we even need the whole showPopupOnHover functionality? I don't think we use it anywhere

oterral: If the `showPopupOnHover` property is not used anywhere, we can remove the functionnality, I guess. But check on the gitlab project too  before you do it.

oterral (approved):
LGTM . Nice fix. it fixes basically all the feature display when you click on a result of the search .

oterral (approved):

PR #1198, opened 2024-10-23
fix(min/max zoom): add minimum zoom to most topics, remove default props from function components

danji90:
# How to
Main feature: Add min zoom to topics:

- Geltungsbereiche
- Ausbildungen Personenverkehr
- Personen mit eingeschränkter Mobilität
- Infrastrukturbetreiberinnen
- Koordinatoren Bahnnahes Bauen
- Verantwortlichkeiten Netzentwicklung
- Beleuchtung Bahnhofklassen
- Netzkarte Energie

Add maxExtent to Tarifverbundkarte

Other updates:
- Remove Intervention SBB related code (very old topic, unused for years)
- Remove defaultProps from function components, use js default values instead

# Others

- [ ] Tests added.


src/components/Dialog/Dialog.js
```diff
@@ -117,27 +117,17 @@ const propTypes = {
...
+    title = null,
     name,
-    isModal,
-    className,
-    onClose,
-    classes: classesProp,
+    isModal = false,
+    className = null,
+    onClose = null,
+    classes: classesProp = {},
```
oterral, line 130: This make "classes" always a new value everytime, so it will re render the component that use it

src/components/DrawButton/DrawButton.js
```diff
@@ -5,7 +5,7 @@ import { useTranslation } from "react-i18next";
 import IconButton from "@mui/material/IconButton";
 import { ReactComponent as Pencil } from "../../img/pencil.svg";
 
-function DrawButton({ children, onClick, ...buttonProps }) {
+function DrawButton({ children = null, onClick = () => null, ...buttonProps }) {
```
oterral, line 8: Same here, It will be always a new function.

oterral (commented):
lgtm.  Be careful with the defaultProps if the default value is an array or object or a function, it will triggers more rerender because the component will consider it is always a new value.

Ideally, prop shouldn't have default value as object, array or function but we should manage undefined value instead.

src/components/DataLink/DataLink.js
```diff
@@ -48,11 +53,4 @@ DataLink.propTypes = {
   fullWidth: PropTypes.bool,
 };
 
-DataLink.defaultProps = {
-  layer: null,
```
oterral, line 52: I guess defaultProps with null value is useless, maybe don#t set default.

src/components/Dialog/Dialog.js
danji90: ok, I changed it to: 
const classesProp = classes || {};
inside the component

danji90:
> lgtm. Be careful with the defaultProps if the default value is an array or object or a function, it will triggers more rerender because the component will consider it is always a new value.
> 
> Ideally, prop shouldn't have default value as object, array or function but we should manage undefined value instead.

This would be prop change, not a state change, which should't trigger a rerender. I cleaned it anyway

PR #1207, opened 2025-01-14
fix(kilometrage): add support for multiple lines in popup

danji90:
# How to

# Others

- [ ] Tests added.


.env
```diff
@@ -6,7 +6,8 @@ REACT_APP_MATOMO_URL_BASE=https://analytics.geops.de/
 REACT_APP_MATOMO_SITE_ID=9
 REACT_APP_VECTOR_TILES_URL=https://maps.geops.io
 #REACT_APP_STYLE_REVIEW_PREFIX=
-REACT_APP_SEARCH_URL=https://maps.trafimage.ch
+# REACT_APP_SEARCH_URL=https://maps.trafimage.ch
```
danji90, line 9: Revert once backend is merged and deployed, do not merge this

src/layers/KilometrageLayer/KilometrageLayer.js
```diff
@@ -8,40 +9,70 @@ import { Feature } from "ol";
...
+      const lines = features.reduce((all, current) => {
+        const lineNumber = current.get("line_number");
+        return lineNumber && !all.includes(lineNumber)
+          ? [...all, lineNumber]
+          : all;
+      }, []);
+      const generalization = features
+        .find((feat) => feat.get("line_number"))
+        ?.get("generalization_level");
+      this.searchUrl = "https://wkp.dev.trafimage.geops.ch";
```
danji90, line 45: Remove before merge and once backend is deployed on prod

src/components/Map/Map.js
```diff
@@ -176,34 +176,36 @@ class Map extends PureComponent {
...
-        }
-      });
-
-      map.getTarget().style.cursor = infos.find(
-        ({ features }) => !!features.length,
-      )
-        ? "pointer"
-        : "auto";
-    });
+    getFeatureInfoAtCoordinate(coordinate, queryableLayers, "pointermove").then(
```
oterral, line 179: this third parameter seems a bad idea. can we avoid it ?

oterral: I propose to do the fetchKilometrage call inside the Kilometrage popup. What do you think ?

oterral (changes requested):
Just one minor change to think about.

src/components/Map/Map.js
danji90: I also thought about that, I liked the idea of having the Popup just handle the UI and letting the layer handle the data. Is there a benefit from doing it in the popup?

And why is the eventType Parameter bad? I thought it might be useful also for other use cases and was surprised we didn't already have a way to access the event. Maybe we could pass in the whole event, not just the type

oterral: if you want a hell of a mess yes. If you start doing this  you will lose sight on what does the getFeatureInfoAtCoordinates , you will have to check every layer every time to be sure what's happening. 

What you could do is renaming the parameter: 'loadExternalData`  true or false, default to true. and if true, this parameter will append some extenal data to the feature your request when you want it .    

Ideally we should have a hasFeatureInfoAtCoordinate(coordiante, layers) on  layers. And this function will only test if there a clickable feature for this layer at the coordinate. Can you do it  ?

